PR #21, opened 2017-07-12
syntax highlighting in readme

petrbel:
An alternative code block syntax is used in order to enable python highlighting of the example. This improves the readability of the example.


michaelschaarschmidt:
Thanks!

PR #27, opened 2017-07-13
Opt out of defaults for Configuration

trickmeyer:
These changes allow a user to create a Configuration to not allow defaults by passing in the "allow_defaults" parameter, which defaults to True to keep the rest of the code working just fine.  This can be a key in a JSON file grabbed by the .from_json() method, or directly as a kwarg into the creation of a Configuration.  In the event it's in both, a consistency check is performed.

It will raise a TensorForce exception if a user creates a Configuration that won't allow defaults and a default is attempted to be set on a parameter not already in the configuration.  Your logic of ignoring the default if the attribute exists still works, and won't raise the exception.

I included unit tests, but I wasn't sure your conventions, so rearrange as you see fit.

This flag could be added to the examples going forward if you wanted to make sure that all keys are always up to date.


trickmeyer:
Checks failing are due to agents not succeeding in tasks, nothing to do with this PR.  It'd be nice to make things a bit more deterministic to avoid these false alarms.

michaelschaarschmidt:
Thanks for the PR, can review tomorrow. Making tests deterministic is difficult because we found seeds to be always working on our machines not working on Travis.

PR #28, opened 2017-07-14
initialize internal state at Agent initialization

nagachika:
I ran into the following error with the example shown in the [blog post](https://reinforce.io/blog/introduction-to-tensorforce/) with DQNAgent.

```
  File "/Users/nagachika/work/tensorforce/tensorforce/agents/agent.py", line 178, in act
    self.current_internal = self.next_internal
AttributeError: 'DQNAgent' object has no attribute 'next_internal'
```

### DIFF
--- a/tensorforce/agents/agent.py
+++ b/tensorforce/agents/agent.py
@@ -144,6 +144,7 @@ class Agent(object):
 
         self.episode = 0
         self.timestep = 0
+        self.next_internal = self.current_internal = self.model.reset()
 
     def __str__(self):
         return str(self.__class__.name)


michaelschaarschmidt:
Thanks for the pr! 
This is what the agent.reset() call is for - the reason this needs to be explicitly called is that one call is necessary in any case between each episode, so the init call would only save one call before the first episode:
https://github.com/reinforceio/tensorforce/blob/master/tensorforce/execution/runner.py#L126

AlexKuhnle:
I guess we could reset the agent from initialization on (we should then also add the preprocessing reset). Currently, this works fine, note however what Michael said about the resets between episodes.

nagachika:
Thank you for your reviewing!
I've noticed that the Agent.reset() should be called between episodes.
In the example on README.md the section [Create and use agents](https://github.com/reinforceio/tensorforce#create-and-use-agents), Agent.act() is called without reset().
I don't mind call reset() before act(). If this is the right way, I'd like to add `agent.reset()` in the example in  README.md.

michaelschaarschmidt:
So the readme example is not a full episode. Reset does not need to be called before act(), just after you finish an episode (see the runner). If you never finish your episode, there is no need to call reset, and not every environment is episodic. We might change the readme example to include a loop and an actual episode to show how reset is used. Thanks for raising the issue.

nagachika:
Thank you for your clarification! I understand it.

Anyway, this is a great project as entry point for RL for the beginners like me. Thank you for publishing TensorForce!

michaelschaarschmidt:
Glad you like it, let us know if you run into any more issues!

PR #38, opened 2017-07-17
Allow custom distributions in policy gradients

befelix:
Not sure if this is consistent with other parts of the codebase, but this is how I currently provide custom action distributions for policy gradients:

```actions = dict(continuous=True, distribution=Gaussian())```

where `Gaussian()` can be replaced by any instance of `Distribution`.


AlexKuhnle:
Thanks @befelix for the PR! I added it and will make it consistent to the codebase.

PR #47, opened 2017-07-24
Maze Explorer environment

mryellow:
It probably works better via OpenAIGym as `gym.wrappers.Monitor` provides periodic video saving features. Though watching it play I did spot a collision bug in the environment itself.

Pull if you think it adds something, might be a half decent integration example though is already heavily based on OpenAIGym example.

### DIFF
--- a/examples/maze_explorer.py
+++ b/examples/maze_explorer.py
@@ -1,0 +1,119 @@
+# Copyright 2017 reinforce.io. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


michaelschaarschmidt:
Hey, thanks for this, we will merge this in the coming days but probably slightly restructure the environments and move them into a separate top-level folder, currently discussing this, hence the holdup

mryellow:
No rush... I haven't tagged first version and registed with PyPi yet.

> probably slightly restructure the environments 

I found the semantics a little tough, maybe a suffix like `OpenAIGymEnv` to differ from the actual module `OpenAIGym`. `MazeExplorerEnv` vs `MazeExplorer`.

Issue #53, opened 2017-07-25
Documentation for epsilon decay


mryellow:
It's not the linear decay based on the remaining I was expecting.

`self.epsilon -= ((self.epsilon - self.epsilon_final) / self.epsilon_timesteps) * timestep`

So over 100 steps that takes about 30-40 to get "close" to `epsilon_final`.

Potential for optional decay mode.

mryellow:
This is an implementation of what I was expecting:

https://github.com/karpathy/convnetjs/blob/9a3fc7be4d47871497a95dd357e52c9190b2c06d/build/deepqlearn.js#L183

In python:

`self.epsilon = min(1.0, max(self.epsilon_final, 1.0-(timestep - self.first_update)/(self.epsilon_timesteps - self.first_update))); `

Though `first_update` isn't defined under `exploration` key of config.

krfricke:
Thanks @mryellow for closing this with #56

PR #56, opened 2017-07-25
Epsilon annealing

mryellow:
See #53

Matching Karpathy's implementation.

https://github.com/karpathy/convnetjs/blob/9a3fc7be4d47871497a95dd357e52c9190b2c06d/build/deepqlearn.js#L183

### DIFF
--- a/tensorforce/core/explorations/epsilon_anneal.py
+++ b/tensorforce/core/explorations/epsilon_anneal.py
@@ -16,10 +16,9 @@
 from tensorforce.core.explorations import Exploration
 
 
-class EpsilonDecay(Exploration):
+class EpsilonAnneal(Exploration):
     """
-    Exponentially decaying epsilon parameter based on ratio of
-    difference between current and final epsilon to total timesteps.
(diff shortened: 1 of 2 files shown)


tensorforce/core/explorations/epsilon_anneal.py
```diff
@@ -0,0 +1,37 @@
...
+    Annealing epsilon parameter based on ratio of current timestep to total timesteps.
+    """
+
+    def __init__(self, epsilon=1.0, epsilon_final=0.1, epsilon_timesteps=10000):
+        self.epsilon = epsilon
+        self.epsilon_final = epsilon_final
+        self.epsilon_timesteps = epsilon_timesteps
+
+    def __call__(self, episode=0, timestep=0):
+        # TODO: Trim by length of `first_update`, removing steps with no learning.
```
mryellow: This can probably be ignored, though with large values of `first_update` it may have an impact.

krfricke:
Thanks, @mryellow !

PR #59, opened 2017-07-26
Shuffle samples to generate each mini-batch at Replay.get_batch.

nagachika:
Previous implementation randomize only the offset in the memory and the mini-batch contains a series of samples in original experienced order.
I believe the experiences should be sampled at random to get rid of bias of experiences.

I've realized `np.random.permutation` with whole memory capacity is not memory efficient. Because I'm newbee on Python and numpy, I cannot find the equivalent way to do this more efficiently. Any hint or suggestions are welcome.

### DIFF
--- a/tensorforce/core/memories/replay.py
+++ b/tensorforce/core/memories/replay.py
@@ -21,7 +21,6 @@ from __future__ import absolute_import
 from __future__ import print_function
 from __future__ import division
 
-from random import randrange
 from six.moves import xrange
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


michaelschaarschmidt:
Hi,

thanks for helping! These are 2 separate strategies and both can be desired. One is to sample each entry separately (not with np.permutaton but with np.random.shuffle and re-assignment of keys), one is to sample continuous ranges.  Hence, I wont merge this but rather suggest you create an issue and then we will see how we can make this optional.

nagachika:
Thank you for your comment!
I didn't know that there's the case that the continuous experiences are desirable. I will file new issue to discuss optional strategies.

PR #63, opened 2017-07-27
Add ALE environment

Islandman93:
Adds the Arcade learning environment from https://github.com/mgbellemare/Arcade-Learning-Environment.

I'm not sure where to put reward clipping (if it should be handled by the agent or the environment) so I just put it here.
Additionally the loss of life termination that was used in the Nature DQN is available as a parameter or a negative reward can be set for losing a life.
While frame_skip is the same as runner repeat actions, it's handled by the underlying code of the ALE and is faster than looping in python.

### DIFF
--- a/examples/ale.py
+++ b/examples/ale.py
@@ -1,0 +1,128 @@
+# Copyright 2017 reinforce.io. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


michaelschaarschmidt:
Hey, thanks for this, we will merge this with the other environment pull request, we have to do some restructuring first

PR #72, opened 2017-07-28
Sequential replay terminals in last index only

### LINKED ISSUE #69: Handle next state semantics in case of random sampling
DQN/DQFD/NAF assume continuous sequences of states, thus do not need to explicitly handle next states, but as we introduced random sampling, this needs to be taken care of. Turned random sampling off by default

Islandman93:
Partly fixes #69


AlexKuhnle:
@Islandman93: Is this obsolete, now that the random sampling next_state thing is solved and terminals don't make problems?

Islandman93:
Yes the next_state fixes this

PR #79, opened 2017-08-01
Add divide by constant preprocessing

Islandman93:
Preprocessing to divide by a constant. Useful if images are returned as uint8 and need to be normalized into the 0-1 range


michaelschaarschmidt:
Thanks for this..will look to add a test soon

PR #86, opened 2017-08-07
Initial TensorFlow summaries support

Islandman93:
Adds tf_summary_level and tf_summary_interval to configs.
Adds a timestep in model that's incremented each call to get_action
Supports loss, trainable variables and action outputs.

The summary levels I had imagined are as follows:

0. Loss/Loss per action, these summaries are always created so they don't check config.tf_summary_level
1. Network Output, Q values for DQN. I'm not very familiar with policy gradient methods so I wasn't sure what output summary would be most useful and thus didn't write any
2. Trainable Variables, target networks is why I had to add a trainable flag into the layer constructor. The easy way to do this was using tf.trainable_variables, but it could be done in the layer functions and that would remove the need for the trainable flag.


8enmann:
Update model.tf_summary argument documentation? Currently says bool

PR #87, opened 2017-08-11
Threaded Asynchronous support for NStep DQN

Islandman93:
- This relies on the tensorflow summaries changes because it makes use of the model.timestep to determine when to update the target network. As a bonus this removes a lot of duplicate code from various Q agents. 
- Working support for NStep DQN model/agent though it doesn't support 2d action arrays. 
- Generic threaded runner.

### DIFF
--- a/examples/configs/dqn_2013_network_visual.json
+++ b/examples/configs/dqn_2013_network_visual.json
@@ -1,0 +1,26 @@
+[
+    {
+        "type": "conv2d",
+        "size": 16,
+        "window": 8,
+        "stride": 4,
+        "bias": true,
+        "padding": "VALID"
+    },
+    {
+        "type": "conv2d",
+        "size": 32,
+        "window": 4,
+        "stride": 2,
+        "bias": true,
+        "padding": "VALID"
+    },
+    {
+        "type": "flatten"
+    },
+    {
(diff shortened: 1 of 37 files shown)


Islandman93:
Looks like a merge conflict in cumulative rewards

The correct code is:
```python
def cumulative_discount(rewards, terminals, discount, cumulative_start=0.0):
    """
    Compute cumulative discounts.
    Args:
        rewards: Rewards from a rollout
        terminals: Booleans indicating terminal states
        discount: Discount factor

    Returns:

    """
    if discount == 0.0:
        return rewards
    cumulative = cumulative_start
    discounted_rewards = np.zeros(len(rewards))

    for n, (reward, terminal) in reversed(list(enumerate(zip(rewards, terminals)))):
        if terminal:
            cumulative = 0.0
        cumulative = reward + cumulative * discount
        discounted_rewards[n] = cumulative

    return discounted_rewards
```

michaelschaarschmidt:
Merged conflicts, but there seems to be a bug here:

        # Synchronise target with training network
>       self.possible_update_target(force=True)
E       TypeError: possible_update_target() got an unexpected keyword argument 'force'
tensorforce/models/categorical_dqn_model.py:62: TypeError

PR #88, opened 2017-08-12
LSTM regression test

mryellow:
Offending commit: https://github.com/reinforceio/tensorforce/commit/b0fd7d13ff1392c60dfef432bfea7a7dd035f71f


mryellow:
Something like this could be added to other agents, or a new test created which imports and reconfigures layout on tests.

PR #99, opened 2017-08-19
Episode reward summary in tensorboard

Islandman93:
Adds saving episode rewards to tensorboard (if using a summary writer)


michaelschaarschmidt:
Seems like there is a conflict now, which way did you intend it?

Islandman93:
Not sure why git can't figure that one out it's:
```python
self.agent.observe_episode_reward(episode_reward)
time_passed = time.time() - episode_start_time
```

PR #102, opened 2017-08-19
Add steps per second to runner & examples

Islandman93:
Adds calculation of number of timesteps per second to examples (and runner) so it's easy to see algorithm speed


michaelschaarschmidt:
Thanks for all the PRs today!

PR #104, opened 2017-08-19
Add layer activations to summaries gteq level 3

Islandman93:
For dense, conv2d, lstm layers


michaelschaarschmidt:
Thanks! Just for me to keep track, what's the current rational for all the different summary levels?

Islandman93:
It was
0: Loss (always created so no need to check summary_level)
1: Output activations, Qvalues or action logits (still need to do this in distributions)
2: Trainable Variables
3: Intermediate layer activations

Though I don't really like levels anymore. A better system IMO is more like a list of the summaries you want like ["loss", "trainable_variables", "output_activations", "intermediate_activations"] but that's a PR for a different time

michaelschaarschmidt:
Ok yeah, there is probably some room to discuss this but can merge this now

PR #117, opened 2017-09-02
Add reward preprocessing to base Agent

Islandman93:
Reward clipping applies to all DQN/A3C agents that I know of so I think it makes sense to put this in the base Agent class. This change supports any type of reward preprocessing as defined in the preprocessing folder (divide, clip [newly added], etc)

Subclasses of Agent must now override _observe(self, reward, terminal) instead of observe. 

Also removes reward clipping from ALE and updates the DQN config files to include reward clipping [-1, 1]

### DIFF
--- a/examples/ale.py
+++ b/examples/ale.py
@@ -42,7 +42,6 @@ def main():
     parser.add_argument('-c', '--agent-config', help="Agent configuration file")
     parser.add_argument('-n', '--network-config', help="Network configuration file")
     parser.add_argument('-fs', '--frame-skip', help="Number of frames to repeat action", type=int, default=1)
-    parser.add_argument('-rc', '--reward-clipping', help="Reward clipping. EX: -1 1", nargs="+", type=float)
     parser.add_argument('-rap', '--repeat-action-probability', help="Repeat action probability", type=float, default=0.0)
     parser.add_argument('-lolt', '--loss-of-life-termination', help="Loss of life counts as terminal state", action='store_true')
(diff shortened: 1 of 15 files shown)


AlexKuhnle:
Hi @Islandman93, thanks for the PR, however, would you mind changing a small thing? Since we do not really use underscore methods at all, could you instead do the same via a super call. If you call `super(...)` at the beginning of each `observe(...)` and then modify (in-place) the list/array of rewards in `Agent.observe(...)`, that should have the same effect, I think. It might be we have to think of a better solution at some point, but I would like to avoid starting with underscore function names now. That would be great, thanks!

Islandman93:
@AlexKuhnle this could be accomplished with a super call like:

```python
# Base Agent
def observe(self, reward, terminal):
    return self.preprocess_reward(reward, terminal)

# Subclass
def observe(self, reward, terminal):
    reward, terminal = super().observe(reward, terminal)
```

As far as I can tell observe is always called with a single reward/terminal. Since these could be python primitives (float, bool) there is no way to modify in place these values. So the above solution (returning the processed reward/terminal) is the only one that will work. 

If you're fine with this solution I'll go ahead and make the modifications

AlexKuhnle:
You're right, it can and usually are value and not reference types (in the case of rewards)... I was somehow thinking about entire batches. If you don't mind changing this, that would be great, otherwise we can also do it. Many thanks!

Islandman93:
Fixed, let me know what you think

AlexKuhnle:
Great, thanks @Islandman93!

PR #121, opened 2017-09-08
Add comment on how to run openai_async.py example

lefnire:
Alex Kuhnle [mentioned on reddit](https://www.reddit.com/r/MachineLearning/comments/6mnp9d/ptensorforce_a_tensorflow_library_for_applied/dk9659h/) how to run this sample. Couldn't find similar documentation in the codebase, found it very helpful so copy/pasting here

### DIFF
--- a/examples/openai_gym_async.py
+++ b/examples/openai_gym_async.py
@@ -15,6 +15,15 @@
 
 """
 OpenAI gym execution
+
+To run this script with 3 workers:
+$ python examples/openai_gym_async.py Pong-ram-v0 -a VPGAgent -c examples/configs/vpg_agent.json -n examples/configs/vpg_network.json -w 3 -D
+
+You can check what the workers are doing:
+$ tmux a -t openai_async  # `ctrl+b d` to exit tmux
+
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Definitely a good idea to add this as a comment to the code. Once we consider the distributed feature as working properly and integrating well with all the models it's supposed to work with, we will also add it to the README etc.

lefnire:
>Once we consider the distributed feature as working properly

Out of curiosity, does the [cluster_spec -based version](https://github.com/reinforceio/tensorforce/blob/master/examples/openai_gym_async.py) function currently? It runs on my local, but I'm not sure if it's doing the A3C thing it's supposed to (trying PPO). I see in threaded_ale.py multiple agents setting their model to a master model, which how I presume it achieves asynchrony. I see the cluster-spec version calling [self.agent.model.set_session(session)](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/execution/runner.py#L118) and have a hunch that achieves the same goal, but I'm quite new to TensorFlow so thought I'd double check.

PR #122, opened 2017-09-08
allow_defaults fix in Configuration.from_json()

trickmeyer:
this method was always allowing defaults rather than passing through the method's argument.


AlexKuhnle:
Thanks for spotting this.

PR #123, opened 2017-09-09
For consideration: add Nadam SGD optimizer

lefnire:
(Close at your digression) Seems like lotta love for Nadam optimizer recently, which is kinda Adam+ as I understand it (Adam +"Nesterov accelerated gradient", [overview](http://ruder.io/optimizing-gradient-descent/index.html#nadam)). With Adam being the default/recommended in many of these agents, might be a boost.


michaelschaarschmidt:
Thanks!

PR #130, opened 2017-09-19
Added to_json() and as_dict() Configuration helpers for easier output of Configurations.

trickmeyer:
Now we can write a modified Configuration to disk and when it loads it's equal (via one.as_dict() == other.as_dict()) to the original object.


michaelschaarschmidt:
Thanks!

PR #141, opened 2017-09-27
Fixing Huber Loss; Config json IO helper methods.

trickmeyer:


AlexKuhnle:
Many thanks, @trickmeyer !

Issue #138, opened 2017-09-25
Replay.get_batch may get wrong next_states when sample the last observation


bigtreeliujie:
If the last observation's index is sampled, as there is no correct next_states in the memory, the result is wrong.

AlexKuhnle:
You're right. If you could create a pull request with the necessary corrections, that would be great. :-)

bigtreeliujie:
OK，I will create a pull request later.

AlexKuhnle:
Closed, thanks for the PR!

PR #142, opened 2017-09-28
Fix bug in replay.get_batch (#138)

bigtreeliujie:
Fix the bug that replay.get_batch may get wrong next_states when sample the last observation ( #138 )


AlexKuhnle:
Many thanks, @bigtreeliujie !

bigtreeliujie:
But, the CI fails...

AlexKuhnle:
That's unrelated, though... our tests are unfortunately currently not working properly with all agents, but the reason why it didn't pass was not because of the memory.

Issue #113, opened 2017-09-01
Distributed NAF agent fails to initialize


admcl:
I'm getting an Attribute error while trying to configure my NAF agent:
```
nafConfig = Configuration.from_json('naf_agent.json') #Copied from examples.

net = layered_network_builder([dict(type='dense', size=512, activation='tanh'), \
                               dict(type='dense', size=512, activation='tanh'), \
                               dict(type='dense', size=512, activation='sigmoid') \
    ])

nafConfig.default(dict(states=environment.states, actions=environment.actions, network=net))

nafConfig.default(dict(distributed=True, cluster_spec=cluster_spec, global_model=(rank == 0), device=('/job:ps' if rank == 0 else '/job:worker/task:{}/cpu:0'.format(rank))))

agent = NAFAgent(config=nafConfig)
```
...Produces the following problem on each node:
```
Traceback (most recent call last):
  File "/home/admcl/NAF_async.py", line 113, in <module>
    agent = NAFAgent(config=nafConfig)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/agents/memory_agent.py", line 62, in __init__
    super(MemoryAgent, self).__init__(config, model)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/agents/agent.py", line 156, in __init__
    self.model = self.__class__.model(config)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/models/naf_model.py", line 45, in __init__
    super(NAFModel, self).__init__(config)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/models/q_model.py", line 48, in __init__
    super(QModel, self).__init__(config)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/models/model.py", line 96, in __init__
    self.global_model = self.__class__(config=global_config)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/models/naf_model.py", line 45, in __init__
    super(NAFModel, self).__init__(config)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/models/q_model.py", line 51, in __init__
    self.possible_update_target(force=True)
  File "/net/fs01.cluster.com/home/admcl/tensorforce/tensorforce/models/q_model.py", line 173, in possible_update_target
    self.session.run(self.target_network_update)
AttributeError: 'NoneType' object has no attribute 'run'
```

[...]

Issue #120, opened 2017-09-07
Async w/ DQNAgent: 'NoneType' object has no attribute 'run' (self.session)


lefnire:
When I run `python examples/openai_gym_async.py Pong-ram-v0 -a VPGAgent -c examples/configs/vpg_agent.json -n examples/configs/vpg_network.json -w 3 -D` and tmux in with `tmux a -t openai_async`, all's running smoothly. If I use DQNAgent configs instead:

`python examples/openai_gym_async.py Pong-ram-v0 -a DQNAgent -c examples/configs/dqn_agent.json -n examples/configs/dqn_network.json -w 3 -D`

then I get 

```bash
CUDA_VISIBLE_DEVICES= /home/lefnire/anaconda3/envs/btc3/bin/python /home/lefnire/Sites/btc/tmp/tensorforce/examples/openai_gym_async.py Pong-ram-v0 --is-child --agent DQNAgent --agent-config /home/lefnire/Sites/btc/tmp/tensorforce/examples/configs/dqn_agent.json --network-config /home/lefnire/Sites/btc/tmp/tensorforce/examples/configs/dqn_network.json --num-workers 3 --task-index -1 --debug
(btc3) lefnire@lefnire-ubuntu:~/Sites/btc/tmp/tensorforce$ CUDA_VISIBLE_DEVICES= /home/lefnire/anaconda3/envs/btc3/bin/python /home/lefnire/Sites/btc/tmp/tensorforce/examples/openai_gym_async.py Pong-ram-v0 --is-child --agent DQNAgent --agent-config /home/lefnire/Sites/btc/tmp/tensorforce/examples/configs/dqn_agent.json --network-config /home/lefnire/Sites/btc/tmp/tensorforce/examples/configs/dqn_network.json --num-workers 3 --task-index -1 --debug
[2017-09-07 13:28:12,300] Making new env: Pong-ram-v0
Traceback (most recent call last):
  File "/home/lefnire/Sites/btc/tmp/tensorforce/examples/openai_gym_async.py", line 182, in <module>
    main()
  File "/home/lefnire/Sites/btc/tmp/tensorforce/examples/openai_gym_async.py", line 151, in main
    agent = agents[args.agent](config=agent_config)
  File "/home/lefnire/Sites/btc/tmp/tensorforce/tensorforce/agents/memory_agent.py", line 62, in __init__
    super(MemoryAgent, self).__init__(config, model)
[...]

PR #148, opened 2017-09-30
[DNM] Distributed fixes

Islandman93:
This changes update distributed to support tensorflow summaries, fixes some important issues,  and allows q_model target networks to be updated in distributed mode.

## Tensorflow Summaries
The "chief" agent (task index 0) will now output summaries in distributed mode. I had to add a supervisor argument into the set_session method on model.py because summaries can only be written through the supervisor in distributed mode. I've modified the distributed example to remove summaries from all workers except task 0.

## Distributed Fixes

1. Global shared optimizer, according to the A3C paper using a shared optimizer produces better results than each worker having it's own. This might make workers like 0.001% and save some memory since they don't have to calculate the optimizer update or store the params just the gradients.
2. Updating local vars. Fixed local vars were being updated from a cached copy of the global variables. Specifically, var.read_value() must be used to ensure tensorflow doesn't used the cached copy. Additionally, local vars are only updated after the gradient calculation. Reset now causes a copy from global -> local, this fixes problems where workers that start slowly had really old parameters. Or if a worker needs to be restarted it will copy from global first.
3. Added general helper methods that allow models to override how the global_timestep is incremented. This allows say DQN to only increment the timestep once even if batch size is 32. 

## Q Model target
Was able to get a tensorflow conditional statement that determines if the global model needs to update the target network. This was the major issue preventing distributed running of DQN & variants. So now all DQN methods should work in a distributed fashion.

I've tested Nstep DQN with these changes and it's now capable of running for an epoch or two before gradient explosion (up from about 10k steps to 1.2mil). I think that this is just a hyperparameter issue and that the distributed code is now correct (NStep DQN still doesn't clip gradients like the original paper so one bad update can mess up the whole thing). 

The policy gradient tests are failing because they don't have optimizers? If that's expected behavior then I can remove the check from the code.

### DIFF
--- a/examples/configs/dqn_nstep_agent_visual.json
+++ b/examples/configs/dqn_nstep_agent_visual.json
@@ -32,7 +32,6 @@
   ],
 
   "batch_size": 6,
-  "repeat_update": 1,
   "keep_last": true,
   "target_update_frequency": 10000,
 
@@ -47,7 +46,6 @@
   "tf_summary": null,
   "log_level": "info",
 
-  "gradient_clipping": 40,
   "update_target_weight": 1.0,
   "double_dqn": false,
   "clip_loss": 1.0
--- a/examples/openai_gym_async.py
+++ b/examples/openai_gym_async.py
@@ -157,6 +157,11 @@ def main():
     logger = logging.getLogger(__name__)
     logger.setLevel(log_levels[agent_config.log_level])
 
+    # don't write summaries for tasks other than chief (0)
+    if args.task_index != 0:
+        agent_config.tf_summary = None
+        agent_config.tf_summary_level = -1
+        logger.info("Summaries disabled for agent {}".format(args.task_index))
     agent = agents[args.agent](config=agent_config)
 
     logger.info("Starting distributed agent for OpenAI Gym '{gym_id}'".format(gym_id=args.gym_id))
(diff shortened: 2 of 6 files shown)


Islandman93:
Just to be complete, this fixes #113 and #120 since those were caused by the target update inherited from q_model

AlexKuhnle:
Thanks @Islandman93 for all the fixes! As we mentioned in Gitter (you have probably seen), we focus on getting out the new version in the next days, so might take a week or so until we integrate this PR.

michaelschaarschmidt:
It seems like this is superseded by the new distributed organisation now?

AlexKuhnle:
I'm just about to change 1-2 things w.r.t. the current architecture using TensorFlow's [Supervisor](https://www.tensorflow.org/api_docs/python/tf/train/Supervisor), since I read that it is or will be [deprecated](https://github.com/tensorflow/tensorflow/issues/6263). The new version will use the [SummarySaverHook](https://www.tensorflow.org/api_docs/python/tf/train/SummarySaverHook) in combination with the [MonitoredSession](https://www.tensorflow.org/api_docs/python/tf/train/MonitoredSession). But I expect that using these classes does automatically handle distributed stuff well (that's what they are there for, I think). So I would suggest we test whether everything is working as expected once it's out.

Islandman93:
Since the api has changed since this PR closing in favor of #214

PR #149, opened 2017-10-03
max of 1 to max of epsilon

wassname:
This was using a max of one which led to funny behaviour where it didn't use epsilon [see plot here](https://gist.github.com/wassname/21b0e3c5a7cf97a32d424cd851507fc3). This fixes it.


AlexKuhnle:
Thanks! As with the other PR, it might take a few days until we integrate it because of a major update.

PR #150, opened 2017-10-03
change half_life of 1 to epsilon_timesteps/10

wassname:
If you plot this is has a half life of 1, independent of epsilon_timesteps. If you plot it you see exploration drop to epsilon_final in ~20 steps. 

This changes it to have a configurable amount of half lives per epsilon_timesteps. [Plots of before and after here](https://gist.github.com/wassname/21b0e3c5a7cf97a32d424cd851507fc3)


AlexKuhnle:
Thanks for the PR! Just to let you know that it might take a few days until we integrate it, since we're working towards getting out a major update in these days.

wassname:
No problems, thanks for all your work on the project

PR #154, opened 2017-10-04
Make PPOAgent a MemoryAgent

wassname:
This change lets PPO use prioritised replay memory which makes it
converge faster on cartpolev0.

However, it converges much slower when using replay memory. I made the
default as prioritized_replay but do we want this change?
@michaelschaarschmidt added PPO so it would be good to get his thoughts
and a code review.

The changes also fixes minor things like the docstring and returning losses for
each instance. As well as updates corresponding tests, config, and
examples.

### DIFF
--- a/examples/configs/ppo_cartpole.json
+++ b/examples/configs/ppo_cartpole.json
@@ -1,15 +1,17 @@
 {
   "log_level": "info",
-
+  "memory": {"type":"prioritized_replay"},
+  "update_frequency":256,
+  "first_update":512,
   "batch_size": 4000,
   "discount": 0.99,
   "gae_rewards": false,
   "gae_lambda": 0.97,
-  "normalize_advantage" : false,
+  "normalize_rewards" : false,
   "learning_rate" : 0.001,
   "entropy_penalty" :  0.01,
   "epochs" : 5,
-  "optimizer_batch_size" : 512,
+  "optimizer_batch_size" : 64,
   "loss_clipping" : 0.2,
 
   "baseline": {
--- a/tensorforce/models/ppo_model.py
+++ b/tensorforce/models/ppo_model.py
@@ -24,8 +24,10 @@ from __future__ import division
 
 import tensorflow as tf
 from six.moves import xrange
+import numpy as np
+
 from tensorforce import util
-from tensorforce.core.memories import Replay
+from tensorforce.core.memories import Replay, PrioritizedReplay, Memory
 from tensorforce.models import PolicyGradientModel
 
 
@@ -36,8 +38,8 @@ class PPOModel(PolicyGradientModel):
     default_config = dict(
         entropy_penalty=0.01,
         loss_clipping=0.2,  # Trust region clipping
-        epochs=10,  # Number of training epochs for SGD,
-        optimizer_batch_size=128,  # Batch size for optimiser
+        epochs=10,  # Number of training epochs for optimizer,
+        optimizer_batch_size=64,  # Batch size for optimizer
         random_sampling=True  # Sampling strategy for replay memory
     )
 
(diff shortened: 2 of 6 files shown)


examples/configs/ppo_cartpole.json
```diff
@@ -1,15 +1,17 @@
...
-
+  "memory": {"type":"prioritized_replay"},
+  "update_frequency":256,
+  "first_update":512,
   "batch_size": 4000,
   "discount": 0.99,
   "gae_rewards": false,
   "gae_lambda": 0.97,
-  "normalize_advantage" : false,
+  "normalize_rewards" : false,
```
wassname: I think this has changes name in the config

examples/configs/ppo_cartpole.json
```diff
@@ -1,15 +1,17 @@
...
   "learning_rate" : 0.001,
   "entropy_penalty" :  0.01,
   "epochs" : 5,
-  "optimizer_batch_size" : 512,
+  "optimizer_batch_size" : 64,
```
wassname: value from the ppo paper

tensorforce/models/ppo_model.py
```diff
@@ -36,23 +38,24 @@ class PPOModel(PolicyGradientModel):
     default_config = dict(
         entropy_penalty=0.01,
         loss_clipping=0.2,  # Trust region clipping
-        epochs=10,  # Number of training epochs for SGD,
-        optimizer_batch_size=128,  # Batch size for optimiser
+        epochs=10,  # Number of training epochs for optimizer,
+        optimizer_batch_size=64,  # Batch size for optimizer
```
wassname: value from ppo paper

tensorforce/models/ppo_model.py
```diff
@@ -36,23 +38,24 @@ class PPOModel(PolicyGradientModel):
     default_config = dict(
         entropy_penalty=0.01,
         loss_clipping=0.2,  # Trust region clipping
-        epochs=10,  # Number of training epochs for SGD,
```
wassname: I think we are using the configure optimiser, not only SGD so I called it the optimiser instead

tensorforce/models/ppo_model.py
```diff
@@ -171,18 +174,45 @@ def update(self, batch):
...
-            elif i == self.updates - 1:  # Last update, fetch return and diagnostics values
-                fetches = (self.optimize, self.loss, self.loss_per_instance, self.kl_divergence, self.entropy)
-                prev_distribution_tensors = {placeholder: tensor for name, placeholders in self.prev_distribution_tensors.items() for placeholder, tensor in zip(placeholders, prev_distribution_tensors[name])}
-                feed_dict.update(prev_distribution_tensors)
-                loss, loss_per_instance, kl_divergence, entropy = self.session.run(fetches=fetches, feed_dict=feed_dict)[1:]
-
             else:  # Otherwise just optimize
                 self.session.run(fetches=self.optimize, feed_dict=feed_dict)
 
+        # For the last epoch, fetch return and diagnostics values for each instance
```
wassname: We need to return the loss for each instance. Prior to this the agent was only returning the values from one minibatch. You can't do this with random sampling (although random sampling is nice idea with PPO) so for the last epoch I go through sequential mini batches of the whole batch.

Suggestion to refactor this code are welcome.

michaelschaarschmidt:
Hello,

sorry for the delay, this will be reviewed once the new optimizers are merged into master, which will change all core structure.

wassname:
No problems I'm happy to wait, the new optimizers should be fun.

michaelschaarschmidt:
I made the default as prioritized_replay but do we want this change? - prioritised replay is currently not implemented efficiently.

wassname:
Maybe it's best not to use this as default. 

My logic was that with replay memory as default, it will be constantly sampling old, and unimportant data which will make it converge slowly With my PR to use log(n) sorting prioritized replay isn't noticeably slower and the memories would be in memory either way.

If you think it's best to make the default as replay, I will. What do you think?

AlexKuhnle:
PPO usually collects a batch of data, and then subsamples from it to perform multiple updates. We use a memory for this, however, it still more fundamentally works as a batch-based agent (note `memory_capacity == batch_size`, not a memory-based one. So really your implementation is a new agent type, inspired by PPO. :-)  Anyway, it wouldn't work as a *replacement/improvement* of the existing PPO implementation.

wassname:
Fair enough, I know what you mean.

Maybe if I can show increased performance on tensorforce benchmarks we can consider adding it as PPOMemmory or something.

wassname:
I know what you mean, it doesn't quite fit with the logic of PPO this way. It would make more sense to get mini-batches directly from the memory agent, rather than having two memories.

No papers have shown the performance of PPO with prioritised memory. Maybe if I show increased performance on the tensorforce benchmarks we can consider adding it as PPOMemoryAgent. I'll close it for now though.

AlexKuhnle:
Yes, if it's a 'good' model (no idea, haven't tried), it's definitely worth considering to add it to the agent pool.

Issue #146, opened 2017-09-29
Easier Tensorboard Customization


trickmeyer:
Ideally the user should be able to define what they want to log in Tensorboard.  The current challenge is that the fetches and feed_dict required to get many of the possible values of interest may not be easy to expose.

One approach would be to create a few hooks or callbacks where the user could run the updates themselves with supplied variables.  Generalizing that across models may not be easy, however.

Sorry this isn't more specific, but I know this item is of interest to at least @Islandman93 and me, so I wanted to get it out there for thoughts.

AlexKuhnle:
Hi @trickmeyer (and @Islandman93),
what do you think about the summary handling in dev (and I think this mechanism came up at some point in Gitter): There is a set of "summary tags" like 'total-loss', 'variables', etc which are defined in the config (e.g. [here](https://github.com/reinforceio/tensorforce/blob/dev/tensorforce/agents/vpg_agent.py#L66)), and the implementation can then check whether a certain tag is requested (e.g. [here](https://github.com/reinforceio/tensorforce/blob/dev/tensorforce/models/model.py#L452)). Makes it possible to select only what is of relevance and allows new models to introduce completely new tags (as long as they don't clash with old ones). We haven't (yet) integrated the `SummarySessionWrapper`, but cleaned up the summary handling in `Model` in general. Essentially, we use the TensorFlow `Supervisor` class which we realized seems to internally handle summary writing every X seconds, so if there are no problems arising with using `Supervisor`, I think that might be the best way to handle it.
I close the issue for now, but feel free to re-open if you feel that doesn't quite reflect what you had in mind.

trickmeyer:
I think it's a good idea to see how the new version feels after some use.

I prefer a lot of Tensorboard logging for debugging and introspection, so I personally use a branch where I pass in a tf.writer I create on my own, so I'll se how that plays along with your `Supervisor` plan.

AlexKuhnle:
Yes, definitely. So currently by default the `Supervisor` writes the log into `config.model_directory`, but it also allows to pass your own writer, if that is desirable (could be done via an optional config value). Would there be an advantage vs `Supervisor` internally creating one?

AlexKuhnle:
We don't so far use it very much, so insight into how others are using Tensorboard logging and what the points to consider are is very welcome.

trickmeyer:
I pass one in / expose it for two main reasons:

[...]

PR #160, opened 2017-10-11
SummarySessionWrapper to reduce Tensorboard tf.summary logging overhead

trickmeyer:
The previous workflow for Tensorboard logging was:

1. Check if the model should log
2. If so, append the summaries to fetches
3. Call session.run()
4. If should log, grab the summary results from session.run(), log them, and update the last log step.

SummarySessionWrapper makes this all transparent to the user with a simple interface:

`
with SummarySessionWrapper(self, fetches, feed_dict) as session:
    returns = session.run()
`

Behind the scenes, it checks to see if the model should log, and if so, tags on model.tf_summaries, which are set up when the instance is created.  when SummarySessionWrapper.run() is called, it only returns the original fetches, and hangs on to the tf_summaries for itself.  They are written on `__exit__`, and the last logged index is updated.  

This is also a step in the right direction for #146.

This has been implemented for the Model base class and PPOModel.  I can get to others upon request, otherwise will as I spend more time with them.

### DIFF
--- a/tensorforce/config.py
+++ b/tensorforce/config.py
@@ -107,7 +107,7 @@ class Configuration(object):
         return self._config.keys()
 
     def copy(self):
-        return Configuration(**self._config)
+        return Configuration(allow_defaults=self.allow_defaults, **self._config)
 
     def as_dict(self):
         d = dict()
--- a/tensorforce/core/networks/layers.py
+++ b/tensorforce/core/networks/layers.py
@@ -174,7 +174,7 @@ def linear(x, size, weights=None, bias=True, l2_regularization=0.0, l1_regulariz
             tf.losses.add_loss(l2_regularization * tf.nn.l2_loss(t=weights))
         if l1_regularization > 0.0:
             weight_l1_t = tf.convert_to_tensor(l1_regularization, dtype=tf.float32, name='weight_l1')
-            tf.losses.add_loss(tf.multiply(weight_l1_t, tf.reduce_sum(tf.abs(weights)), name='loss_l1_weights'))
+            tf.losses.add_loss(tf.multiply(weight_l1_t, tf.reduce_sum(tf.abs(weights))))
         else:
             weight_l1_t = None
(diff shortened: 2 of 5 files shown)


trickmeyer:
one minor update needs to be done to not create red herrings when exceptions are raised during the run() method:

`
    def __exit__(self, exc_type, exc_val, exc_tb):
        if exc_tb is None:
            if not self._have_executed:
                raise TensorForceError('{} exiting without run() being called.'.format(self.__class__.__name__))
            if self._should_write_summaries:
                self._model.write_summaries(self._returns[-1])
                self._model.last_summary_step = self._model.timestep
        else:
            return 0
`

AlexKuhnle:
Thanks, @trickmeyer, that sounds very useful, and potentially easier to integrate into the new version. We will come back to you.

trickmeyer:
I went ahead and committed the change that will pass through exceptions encountered during session.run()

trickmeyer:
This can only be used on graph runs where all summary variables are calculated.  If specific summary nodes aren't activated in a particular run this will fail.  We could handle this with a try/except, but it seems like a waste.  Probably worthy of pairing the correct run fetch/feed.

AlexKuhnle:
We moved to using the TensorFlow `Supervisor` class in the new branch for session handling ([see here](https://github.com/reinforceio/tensorforce/blob/dev/tensorforce/models/model.py#L206)), and realized that it seems to implement exactly the functionality of your proposed `SummarySessionWrapper`. I'm not sure whether using `Supervisor` might be 'too much' in some cases, we have to investigate that...

michaelschaarschmidt:
Ok this might have to be closed, summary handling is just handled entirely different now.

trickmeyer:
You got it.

michaelschaarschmidt:
Sorry for the extra work

trickmeyer:
@michaelschaarschmidt none at all.

PR #162, opened 2017-10-13
Fix  'observation' referenced before assignment"

wassname:
If prioritized_replay enters the loop in line 89, it gets the index but not the observation, leading to the error "UnboundLocalError: local variable 'observation' referenced before assignment". This fixes it.


wassname:
This can wait until after your optimiser work, I'm just putting it out there. Cheers

AlexKuhnle:
Thanks, we will look into this probably next week.

PR #163, opened 2017-10-13
error instead of infinite loop

wassname:
If the first update is smaller than the batch size them prioritized_replay can go into an infinite loop on lines 93 and 98, this will raise an informative error instead.


wassname:
This can wait until after your optimiser work, I'm just putting it out there. Cheers

PR #164, opened 2017-10-13
Warning when get_batch can be very slow

wassname:
When entering loop 97, if len(observations) is not much larger than batch, it is extremely slow (minutes), due to the way the algorithm works. This was very confusing to me so I thought it might be valuable to add a warning, although I'm not sure if it's needed and I'm not sure if this is the best way. Happy to go with what you think.

This can wait until the optimiser update, cheers.;

### DIFF
--- a/tensorforce/core/memories/prioritized_replay.py
+++ b/tensorforce/core/memories/prioritized_replay.py
@@ -42,7 +42,6 @@ class PrioritizedReplay(Memory):
         self.last_observation = None  # stores last observation until next_state value is known
         
         self.logger = logging.getLogger(__name__)
-        self.logger.setLevel(util.log_levels[config.log_level])
 
     def add_observation(self, state, action, reward, terminal, internal):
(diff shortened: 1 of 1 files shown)


wassname:
Looks like I was wrong about this, it's just a search algorithm that scales poorly

AlexKuhnle:
It is indeed not a particularly fast implementation, but a naive linear search version where one really could use binary search, a priority queue, or so. We might optimize that at some point, but it definitely won't happen soon. If you're keen to look into it, a PR would be very welcome... :+1:

wassname:
Fair enougth, it's new and not many people have implemented it. Optimisation is a bit much to expect! 

I have a version where I used numpy linear search, it's 10x faster and scales a little better. I'll make it a PR and let you guys decide if that small improvement is worth merging.

michaelschaarschmidt:
This would be welcome, although we would have to re-implement this in TensorFlow soon

PR #168, opened 2017-10-14
use sum tree for O(log(n)) searching

wassname:
This is a faster prioritised learning search alg as suggested in the original paper.

Speed:

|search space|old speed (ms)|numpy (ms)| sum-tree (ms) |
|-|-|-|-|
|1e3|13.9 |0.9| 1.45 |
|1e4|132.0 |4.7| 1.7 |
|1e5|1690.0 |55.6| 2.1 |
|Order|O(n**1.04)|O(n**0.9)|~O(log(n)) or O(n**0.08)|


michaelschaarschmidt:
Hey,

we are looking to merge this now. Some comments:
- We want to implement the memories in TensorFlow as well and it is not clear whether this strategy would be easily implemented there
- One way would be to keep the old prioritised replay and have this one as a separate class first so we can have this version in python if it's inconvenient in TF

wassname:
Good points, and in the end you decide and I'll go with that.

My thoughts are that the old one will be kept around in the older commits, that way it can be used as reference to make the tensorflow version. This implementation might be able to be ported to tensorflow (it revolved around a list of None's and floats, but is not very vectorized). And at that point the tensorflow version would be the best.

But if you decide to keep both for that reason, it's fine with me. Want me to implement that, and fix the conflicts?

AlexKuhnle:
I think it would be best if for now you could rename the old version to `naive_prioritized_replay` and yours as the 'default' `prioritized_replay`. We can then see in the next weeks whether we get rid of it. If you could do that, would be great. :+1:

wassname:
Sounds good.

And thanks for all the work on 0.3.0, it's awesome!

michaelschaarschmidt:
Thanks for implementing!

PR #171, opened 2017-10-18
missing Exploration.from_config

lefnire:
As w/ all my newbie PRs, feel free to scrap mine & use your own if y'all have a correct solution. Defining exploration in the config gave "Exploration doesn't have from_config()" (ish).


AlexKuhnle:
Thanks! The `from_config` is not necessary, instead in other parts it should rightly be called via `from_spec`. Will fix that.

PR #178, opened 2017-10-20
Add Dropout layer, Lstm dropout handling, pass training placeholder during layer.apply()

lefnire:
Attempt#1 for handling dropout. Background as I know it: dropout should be applied as an additional layer (1) after the input layer, (2) after any dense layers. For Lstm layers, dropout is handled internally by TensorFlow w/ a different mechanic, so should be passed as an argument (as it currently is). AFAIK you don't use dropout w/ Conv2d ([to investigate](https://www.reddit.com/r/MachineLearning/comments/42nnpe/why_do_i_never_see_dropout_applied_in/)). So an example: `[{type: 'dropout', rate: .5}, {type: 'lstm', dropout=.5, ...}, {type: 'dense', ...}, {type: 'dropout', rate: .5}]`

The dropout layers (`tf.layers.dropout` & `tf.contrib.rnn.DropoutWrapper`) take a `training` `tf.placeholder` argument. Usually `tf.placeholder_with_default(False, ...)` so we're forced to explicitly activate it during training (that is, during `model.update()`). 

This was simpler in #master (sloppy commit https://github.com/lefnire/tensorforce/commit/4c0224ca719cc8e827029f0ccca6b41846f16187). It's tougher here; as you can see I'm passing the `training` arg to every `network.apply` => `layer.tf_apply`, which results in scattering this arg all over the place. It would be cleaner if `training` were passed to `__init__` functions, but the order in which model subclasses are calling parent `__init__` fns makes it difficult to so do (LMK if you want a detailed explanation of that). 

Anyway, worth chatting about - I expect this won't be merged as is :)

### DIFF
--- a/tensorforce/core/baselines/network_baseline.py
+++ b/tensorforce/core/baselines/network_baseline.py
@@ -46,7 +46,7 @@ class NetworkBaseline(Baseline):
         super(NetworkBaseline, self).__init__(scope, summary_labels)
 
     def tf_predict(self, states):
-        embedding = self.network.apply(x=states)
+        embedding = self.network.apply(x=states, training=False)
         prediction = self.linear.apply(x=embedding)
         return tf.squeeze(input=prediction, axis=1)
 
--- a/tensorforce/core/networks/__init__.py
+++ b/tensorforce/core/networks/__init__.py
@@ -13,12 +13,13 @@
 # limitations under the License.
 # ==============================================================================
 
-from tensorforce.core.networks.layer import Layer, Flatten, Nonlinearity, Linear, Dense, Dueling, Conv1d, Conv2d, Lstm
+from tensorforce.core.networks.layer import Layer, Flatten, Dropout, Nonlinearity, Linear, Dense, Dueling, Conv1d, Conv2d, Lstm
 from tensorforce.core.networks.network import Network, LayerBasedNetwork, LayeredNetwork
 
 
 layers = dict(
     flatten=Flatten,
+    dropout=Dropout,
     nonlinearity=Nonlinearity,
     linear=Linear,
     dense=Dense,
@@ -33,6 +34,7 @@ __all__ = [
     'layers',
     'Layer',
     'Flatten',
+    'Dropout',
     'Nonlinearity',
     'Linear',
     'Dense',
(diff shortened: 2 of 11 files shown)


AlexKuhnle:
Thanks @lefnire. It's a bit like `deterministic` in the case of distributions, I guess. Maybe it's worth handling all this via an `initialization` function which is called for distributions/layers/etc in `Model.initialize`, but we have to think about this. One question though: It seems to me as if the `training` boolean placeholder really is an `update` placeholder (training `act()` still should not apply dropout), so I would probably suggest renaming it (and explicitly setting both values, instead of `placeholder_with_default` for clarity?).

lefnire:
>Maybe it's worth handling all this via an initialization function which is called for distributions/layers/etc in Model.initialize

Agreed. I hit snags trying such in the order in which Model subclasses call super. I'll sit with it again & write a proper description to show what I mean. Will chew on the second part of your comment soon & make necessary adjustments.

michaelschaarschmidt:
Thanks for updating conflicts, will be merged shortly

lefnire:
Snap, I appreciate! Just do a double check that it all makes sense according to the grand scheme if you could. I've been using it in the wild for a while now, seems to work hunky-dory. Just felt a bit hack-job the way I did it.

AlexKuhnle:
Hi @lefnire, just to confirm: The training flag is set only for the update, right? Because for another reason I integrated an update placeholder, and I would then go ahead and combine them (I think the distinction update/non-update is clearer than training/testing in RL, as mentioned above).

lefnire:
@AlexKuhnle you might need to go with your gut on this, my newness to RL strikes again - I still don't quite get the difference. In supervised learning, dropout is applied only during `session.run(minimize_loss)`. Any time you're making a prediction (even if checking training residuals) the flag is `False`. However that translates :/

AlexKuhnle:
The difference would be that even in the 'training phase' one always wants to act to the best of one's knowledge (hence the `training` flag is set to `False`), even though the same experience is then used in update with dropout (`training` set to `True`) to optimize.

lefnire:
I think that sounds like the right way. Yeah I think you only want to apply dropout during backprop / loss-min, and your way sounds compatible

AlexKuhnle:
I'm changing this then, essentially replacing `training` with `update`, and integrating it a bit more with the library structure. That's definitely a useful flag, I think, if only for more specific things. Thanks again!

The reason why the `update` flag was added, for instance, was because only during an update summaries can be extracted, because only then the full `feed_dict` covering all placeholders is passed. This just fyi. ;-)

PR #193, opened 2017-10-26
Add Dueling Layer and 'none' nonlinearity for future development

Mazecreator:
This pull request introduces 'dueling' layer for testing as well as a none as a nonlinearity to simplify code complexity of RainbowAgent which is under development.

The pull request will recognize the following network:
# Network as list of layers
    network_spec = [
        dict(type='dense', size=32, activation='tanh'),
        dict(type='dense', size=32, activation='tanh'),
        dict(type='dueling', size=10, activation='none')
    ]


AlexKuhnle:
Thanks for the PR, @Mazecreator! It seems to me this dueling layer might actually be more appropriately understood as a `Distribution` implementation, but have to think about this more.

Mazecreator:
I was thinking the same thing as the Paper seems to imply it is the output layer which you say really can't be done from the Network layer.  I am thinking it can stay as a layer but maybe we rename it to something like ValueAdvantage or something like that so it isn't confused as a "Dueling Layer" implementation.

AlexKuhnle:
I think it's fine for now, after all it links to the paper. But it might be we move it at some point to the distributions -- it's something to be thought about a bit more anyway... ;-)

Issue #118, opened 2017-09-03
Idea for baseline


GolovanovSrg:
For baseline usefully use body of main neural network and create layer for value function.

michaelschaarschmidt:
Hi,
thanks for the suggestion, it's something we discussed on the weekend as we updated baselines for multi-state support.

The question is what we consider the body of the main neural network. In the case of DQN with convolutional layers + dense layers, just the convolutions? All of it? How can it be configured from the client side? 

We were still debating on how to do it, do you have any suggestions?

GolovanovSrg:
You can do the same as internal_input and internal_output in the network builder - baseline_output. And by default as baseline_output  you can take network builder output.

bigtreeliujie:
I understand @AlexKuhnle keeping network builder a clean feature. In my option,  this abstraction of network is a good idea and should be insisted during Tensorforce's whole life cycle.

In the other hand, sharing some network layers or even complicated network architectures between baseline network and policy network is also a common scene in RL. Using baseline means there is not only one network in the model. Even worse,  some new algorithm involves more networks. For example, , in [ICM](https://arxiv.org/abs/1705.05363) with AC there is four network: policy network, value network, forward model, inverse model, and some parts of those network are shared among them. 
I do not think the  "core network" is a good idea. Because there may be many "core network": network a and network b share a core, network b and c share another core ... 
As a fundamental library, Tensorfoce should provide users the flexibility to implemented their complicated networks, supporting sharing parts among networks.

Sharing network parts between networks is one of the features of Tensorflow. As a result, the network builder could use this feature in something like below:

```python
def network_part_builder(x, reuse=False):
    with tf.variable_scope(name="core1", reuse=reuse):
        x = tf_fnn(x, ...)
        ...
    return x

def policy_network_builder(x):
    core_output = network_part_builder(x)
    x = fnn(core_output, ...)
    return tf_softmax(x)

def value_network_builder(x):
    core_output = network_part_builder(x)
    return fnn(core_output, output=1, ...)
```

In current version baseline is implemented having its own predict and update interface, both interfaces do tf.session.run. In my option, this is not a good idea, because multiple session.run means several IO between python and Tensorflow. However, the prediction results of baseline getting form Tensorflow are immediately sent back to Tensorflow when update policy network.  
We could do the prediction when update the policy network in one Tensorflow run:
  - create baseline's network using the policy network' placeholders (Model.state, Model.reward) as inputs
[...]

PR #198, opened 2017-10-27
Allows custom baseline network

lefnire:
Eh, "PR".. one-liner.. Re: https://github.com/reinforceio/tensorforce/issues/118#issuecomment-340088616. Handy for people using a custom-defined network, rather than `network_builder` ([example](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/tests/test_tutorial_code.py#L226))


AlexKuhnle:
Thanks! Sometimes even one-liners can add new features... ;-)

PR #212, opened 2017-11-03
Add start of Skip connection (ResNet) for Dense Layer

Mazecreator:
This implements skip connection like ResNet (https://arxiv.org/pdf/1512.03385.pdf) on the dense layer.  The input and output size need to match (for now, maybe configure a weight matrix to remap later).  To accomplish this, size=None and it will have the class Layer use the input as the output size of the weight matrix so it is square.

skip=False by default and is used to enable the skip connection.  Example network might look like this:
```
    network_spec = [
        dict(type='dense', size=32, activation='tanh', skip=False),
        dict(type='dense', size=None, activation='tanh', skip=True),
        dict(type='dense', activation='tanh', skip=True)
    ]   
```

Layer 1 will map inputs to 32 outputs.  Layer 2 and 3 are both Skip Connections (duplicate layers and add skip).  Change should not break current implementations.

Side effect if size is left off, it will use the output of the previous layer as the size of that layer.

I will work on adding the "skip" option to the CONV layers if this looks like a good approach.


AlexKuhnle:
Looks good, thanks -- however, why are there so many commits, and why does e.g. the last one essentially add all lines in core/networks/__init__.py? It is much appreciated if PRs consist of few meaningful commits (in this case the second last contains all the changes).

Mazecreator:
I am not entirely sure, I know most of them were me pulling from your repo back into mine and I was trying to work with the Deuling at the same time.  I am not very good with "git" yet.  Is there a way for me to clean this up?  I know the change is accurate as shown in the files.

michaelschaarschmidt:
I might enable squash and merge so all commits of the Pr will be merged into one commit automatically. Is this ready for merge @AlexKuhnle ?

AlexKuhnle:
There is one print statement that probably shouldn't be there, but apart from that it can be merged. No worries @Mazecreator, happens to me as well, it's best to always pull before doing even any commits locally.

Mazecreator:
Sorry, couldn't figure out how to "rebase" or "reset" the problem.  I have deleted and re-forked my repo from yours to make sure I am in sync again.

PR #215, opened 2017-11-05
fix "object has no attribute 'weights'"

wassname:
This fixes "object has no attribute 'weights'" when trying to set weights using a float on a Linear layer


AlexKuhnle:
Thanks for this, @wassname!

PR #217, opened 2017-11-07
Revise DQN agent example config for new API

ImpulseAdventure:
This should address the errors that occur with existing config.
New keys from `default_config` dict have not been migrated yet.


michaelschaarschmidt:
Thanks!

PR #219, opened 2017-11-07
Fixes to Dueling layer

lefnire:
Some merges walked on each other, here's some small fixes to Dueling


Mazecreator:
LGTM,  I have one question which I may have messed up when I first coded this layer:

https://github.com/reinforceio/tensorforce/blob/master/tensorforce/core/networks/layer.py#L510-L524

I have been burnt with variables replaced by Python rather than copied.  Does Append copy the value in line 519 so then assign in 521 doesn't replace the same value ultimately pushing 2 of the same value onto the regularization stack?

AlexKuhnle:
@Mazecreator, the Python operations are executed greedily and procedurally, hence the append of the referenced object is executed before the reassignment to the new value is done. So should be all fine... :-)

PR #220, opened 2017-11-08
Enhanced error handling for unknown objects

ImpulseAdventure:
Should resolve the `ValueError` on `obj.rsplit()` when an unknown agent name is provided.

Example failure stimulus:
`python examples/openai_gym.py CartPole-v0 -a PPOAgent -c examples/configs/ppo_agent.json -n examples/configs/ppo_network.json` 

In the above case `PPOAgent` is not a valid option (it should be `ppo_agent` in the revised API) and we receive a non-intuitive error message:
`ValueError: not enough values to unpack (expected 2, got 1)`

With this fix, if the object string is not found, it will report a list of valid options.


AlexKuhnle:
Thanks!

PR #221, opened 2017-11-09
Minor: fix VPG LSTM example config

ImpulseAdventure:
Added missing size key for LSTM layer (VPG example config)


michaelschaarschmidt:
Ah we did not use that key because we assume the LSTM per convention to have the size of the previous layer but I suppose we could add it

PR #223, opened 2017-11-09
add summary_spec for 'inputs' show states,actions,rewards

Mazecreator:
This adds histograms for the States, Actions, and reward input values if "inputs" is added to the labels of the summary_specs

screenshot 2017-11-09 16 37 35

### DIFF
--- a/tensorforce/models/model.py
+++ b/tensorforce/models/model.py
@@ -210,6 +210,16 @@ class Model(object):
                     deterministic=self.deterministic_input
                 )
 
+            if 'inputs' in self.summary_labels:
+                for name, state in states.items():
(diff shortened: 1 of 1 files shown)


Mazecreator:
I have an update for this to include "regularization" details, but I can't get the L1/L2 data into my DQN to test it.  I can push it if you would like as it catches the "entropy" but that seems to be the only value returned.

Mazecreator:
I did add "regularization" as well that might help debugging or investigating issues.

Mazecreator:
Found the issue with the missing L2/L1 regularization, new pull request submitted

AlexKuhnle:
Thankt @Mazecreator!

PR #226, opened 2017-11-10
Fix regularization in layer.py

Mazecreator:
Module name was incorrect, was never called by network layers.


michaelschaarschmidt:
Good, spot, thanks!

PR #227, opened 2017-11-10
Fix reset() when gym monitor enabled

ekerazha:
Fix reset() when gym monitor enabled (see: https://github.com/rll/rllab/commit/4fd25fe7a31388fed3e005c86bee9b1c75e4df1d).

### DIFF
--- a/tensorforce/contrib/openai_gym.py
+++ b/tensorforce/contrib/openai_gym.py
@@ -58,6 +58,8 @@ class OpenAIGym(Environment):
         self.gym = None
 
     def reset(self):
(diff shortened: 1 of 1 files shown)


krfricke:
Open AI gyms stats recorder requires the episode (self.done) to be finished before resetting the environment for the next episode (https://github.com/openai/gym/blob/master/gym/monitoring/stats_recorder.py#L67). With this PR circumvent this measure (therefore forfeiting stats data on the last episode). However, since we usually collect episode statistics in the tensorforce runner, I don't see a problem in bypassing this restriction.

PR #233, opened 2017-11-15
Adds summary_spec gradients for TF optimizers

Mazecreator:
Adding "gradients" to the summary_spec label will product histograms for optimization variable gradients.  This will currently only work for TensorFlow variables as it requires TF's "compute_gradients()" function.  Will raise an error if not of TFOptimizer.
screenshot 2017-11-15 17 39 34

### DIFF
--- a/tensorforce/models/model.py
+++ b/tensorforce/models/model.py
@@ -53,7 +53,7 @@ import numpy as np
 import tensorflow as tf
 
 from tensorforce import TensorForceError, util
-from tensorforce.core.optimizers import Optimizer, GlobalOptimizer
+from tensorforce.core.optimizers import Optimizer,TFOptimizer, GlobalOptimizer
 
 
 class Model(object):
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Looks good. Two comments: I would add the summary in the optimizer class instead of accessing it in the model if it exists. However, I realize that currently the necessary arguments are not available in the optimizer. Moreover, the error shouldn't be thrown, since the philosophy behind the labels is meant to be: if it's set and a module recognizes it, then add, else nothing happens. Do you want to look into the first point yourself (similar to how summary_labels are passed to `Network`, `Layer`, `Distribution` etc)? Otherwise we can move things around accordingly at some point...

Mazecreator:
I will take a look, might not be until after Thanksgiving here in the USA.

Mazecreator:
Closed by: https://github.com/reinforceio/tensorforce/pull/250

PR #237, opened 2017-11-18
Add `sess_config` model argument for specifying optional ConfigProt extras

lefnire:
For consideration. This adds an optional `sess_config` to be populated with `tf.ConfigProto()` from the user. Double-check where I'm using it in `model.py`, I just replaced every `config=None` with `config=self.sess_config` - which works on local, but I'm less versed with these more advanced session classes.

For reference, I've got a beefy 1080ti I like to multi-task runs on. Eg, if I'm exploring a new model I'd rather see it in 4 single-runs than a 4-worker threaded-run (I've been told that paints a more accurate picture). So for each single-run, I pass in:

```
config['sess_config'] = tf.ConfigProto(
    gpu_options=tf.GPUOptions(per_process_gpu_memory_fraction=.85/4)
)
# .82 instead of 1 since TF tacks on some general GPU overhead, need wiggle room
```


AlexKuhnle:
Thanks for the PR, it's a config/parameter extension we wanted to integrate at some point anyway, so very helpful. One thing: Could you rename it to `session_config`, since we try to avoid abbreviations (unless names get too big)? Otherwise we can also do it, of course...

lefnire:
Done

AlexKuhnle:
Thanks, @lefnire !

PR #238, opened 2017-11-18
Check self.batched_observe is not None

lefnire:
Looks like a "disabled" `batched_observe` attr is expected to be `None` (rather than 0) the way it's handled elsewhere in `agent.py`. Makes consistent in this location.


michaelschaarschmidt:
Thanks!

PR #241, opened 2017-11-21
Fix keyword argument issue in get_batch() random_sampling

ImpulseAdventure:


michaelschaarschmidt:
Why is that an issue? Keywords seem correct to me?

ImpulseAdventure:
Apparently this is a case wherein the function doesn't take keyword arguments. Using Python 3.6.1 and numpy version 1.13.3, the following error is reported with the existing code, but addressed by converting to positional:
`TypeError: where() takes no keyword arguments`

michaelschaarschmidt:
Ah, was not aware of that, thank you for pointing it out

PR #245, opened 2017-11-25
Fix comment spelling and trailing whitespace.

ngoodger:


AlexKuhnle:
Hey @ngoodger, the second part of the PR, trailing whitespace, is actually on purpose. It signals that the line should not actually break in the final compiled documentation. So if you could undo the second bit, that would be great.

ngoodger:
Hi Alex.  Thanks, good to know.  I have reverted the trailing whitespace commit.

michaelschaarschmidt:
Thanks!

PR #250, opened 2017-12-01
Add TensorBoard TEXT configuration and better implementation of gradient capture

Mazecreator:
This will close https://github.com/reinforceio/tensorforce/pull/233 as there is a better implementation pushing the summary capture into the Optimizer class

Added feature of capturing Metaparameter from "Agent" and populating the TensorBoard TEXT screen with that configuration data for rapid review.  Custom Dictionary data can be include to populate into TensorBoard.

        metaparams['MyDescription'] = "This experiment covers the first test ...."
        metaparams['My2D'] = np.ones((9,9))   # 9x9 matrix of  1.0's
        metaparams['My1D'] = np.ones((9))   # Column of 9   1.0's

        summary_spec=dict(directory=CkptDir, meta_dict=metaparams, steps=50,labels=['print_configuration','configuration','gradients']),      

labels: 
  * print_configuration - outputs configuration data to STDOUT
  * configuration - populates TensorBoard TEXT panel with configuration data
  * gradients - captures gradient data from training

PR #252, opened 2017-12-02
Tensorboard Configuration and Gradient Summary

Mazecreator:
Roll-up of https://github.com/reinforceio/tensorforce/pull/250 changes with function names in lower case with underline.  Resolved conflicts with latest commit.


michaelschaarschmidt:
Thanks for updating again, 2 final things:
- Could you use named arguments instead of **kwargs in the optimizers? This increases readability a lot
- There are still various lines with high length that need to be reformated
'+            raise TensorForceError("Error:  MetaParameterRecorder Dictionary conversion was passed a type {} not supported.".format(str(type(data))))
'

AlexKuhnle:
Repeating comment in other PR:

One additional comment: Could you make the arguments [here](https://github.com/reinforceio/tensorforce/pull/250/commits/6dcc60536df2a046f72faa94b52b5cc2f631a25c#diff-12aa0668d288202fb1bef17c18f6b76aR34) explicit, like e.g. [here](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/core/networks/layer.py#L38), instead of using `**kwargs`? I have to look into [this part](https://github.com/reinforceio/tensorforce/pull/250/commits/6dcc60536df2a046f72faa94b52b5cc2f631a25c#diff-7e3f4e6f7ac8c85dfed66ccf8e458e03R881) as well (summaries are integrated via the summary hook, so shouldn't be necessary to call them explicitly), but can do that after the merge and adapt things if necessary.

Mazecreator:
I will hold off on additional update until you look into the summaries issue.  The reason why I put it there is because you only want these summary ops called once per execution.  Otherwise it will load tensorboard with hundreds of duplicate text copies.  I couldn't find a clean place to put this into the summary_hook where it would only be called once.

I will give the explicit arguments another shot, when I was first trying to implement this I tried explicit arguments and got all sorts of Init errors.  I might have a better understanding of the flow (Think it was multi_step that threw me) and can probably get that cleaner.  I was thinking of adding Gradient Clipping feature as well and didn't want to refactor everything again, but I think I can use the Optimizer parameter to pass type and value of clipping cleanly.

For the long TensorForceError's should I simply cut the line with a "\" and continue on at character 1 of the next line for the text or is there a better way to trim that down to how you would like it?

AlexKuhnle:
Ah, I didn't understand this part about how you handle this summary call, but that sounds perfectly fine then. I would say don't worry about the line break for the over-long string, we can do that for now (and really should get some formatting guidelines out). So if you could look into the kwargs thing, that would be great!

Regarding gradient clipping, this is actually a feature that we just added over the weekend ([here](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/core/optimizers/clipped_step.py)), although I just realized that there is a small bug. FYI, it does not clip gradients, since that's harder for us to integrate and less general (non-gradient-optimizer). We instead clip the resulting step.

AlexKuhnle:
(fixed problem in clipped_step optimizer)

Mazecreator:
Latest push add parameters to the optimizer class with one exception.

Additional control for summary data has been added to help reduce the growth rate of the summary data:

        summary_spec=dict(directory=CkptDir, meta_dict=metaparams, steps=50,labels=['print_configuration','configuration','gradients_scalar',\
            'gradients_histogram','gradients','regularization','inputs','states','actions','rewards']),   

New labels have been added:
  * gradients: all gradients
    *  gradients_scalar: Scalar Mean/Variance of gradients
    *  gradients_histogram: raw gradient output (very large)

  * inputs: all inputs
    * states: States only
    * actions: Actions only
    * rewards: Rewards only

michaelschaarschmidt:
Thanks for the updates, merged, we will see about the meta optimizer kwargs

PR #255, opened 2017-12-05
Update index upon set_memory()

ImpulseAdventure:
I believe the buffer indices will get out of sync in the current implementation should `set_memory()` be called instead of `add_observation()`. This fix has not been fully tested.


michaelschaarschmidt:
Thanks, yes that looks sensible. 

PS: Memories will be moved into TensorFlow next just in case you are working around them.

PR #256, opened 2017-12-06
quickstart guidance re: embedding for gym discrete obs space

ImpulseAdventure:
Minor update: Might be helpful for those using `quickstart.py` to be aware of the need to adapt discrete OpenAI gym observation spaces to the tensor of floats required by existing TensorForce layers. Added brief note but feel free to adjust (eg. drop identification text) as appropriate.


michaelschaarschmidt:
Yes that's a good idea, thank you

PR #258, opened 2017-12-06
Add `should_observe` runner option for train/test scenarios

lefnire:
See conversation  at https://gitter.im/reinforceio/TensorForce?at=5a27fd6f3ae2aa6b3f933b18. I haven't tested / investigated this enough, so double check this is all that's needed - kinda WIP here to kickstart

### DIFF
--- a/tensorforce/execution/runner.py
+++ b/tensorforce/execution/runner.py
@@ -56,7 +56,8 @@ class Runner(object):
         episodes=None,
         max_episode_timesteps=None,
         deterministic=False,
-        episode_finished=None
+        should_observe=True,
+        episode_finished=None,
     ):
         """
         Runs the agent on the environment.
@@ -66,6 +67,7 @@ class Runner(object):
             episodes: Number of episodes
             max_episode_timesteps: Max number of timesteps per episode
             deterministic: Deterministic flag
+            should_observe: Should the agent observe rewards? (set to False if testing a trained agent)
             episode_finished: Function handler taking a `Runner` argument and returning a boolean indicating
                 whether to continue execution. For instance, useful for reporting intermediate performance or
                 integrating termination conditions.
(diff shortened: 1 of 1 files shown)


lefnire:
Note: this causes issues with `Runner.agent.episodes` counter (so it never ends). Investigating

michaelschaarschmidt:
Observe induces a graph call to the episode counter which needs to happen either way.  I am wondering however if this should not go into the benchmark instead if it makes the runner more confusing

lefnire:
Sure, your call - feel free to close. I never noticed the benchmark repo, I'll dive in this weekend

AlexKuhnle:
We could add some basic functionality like this also to the runner, after all, it's only a couple of lines so far. However, the benchmark project is definitely the place for more sophisticated evaluation-while-training schemes (for now, at least). Would you mind changing the parameter name to something like `evaluation=False`, which from an API perspective 'describes' more what it's actually meant for? :+1:

lefnire:
Sure thing - but it sounds like this code isn't the way to go, because `observe()` is required to update the episode count, which in turn will break `run()` - so this code is an infinite loop "fundamentally". My new code looks like:

```
env = ...
agent = ...

n_train, n_test = 250, 50
runner = Runner(agent=agent, environment=env)
runner.run(episodes=n_train)  # train
for i in range(n_test): # test
    next_state, terminal = env.reset(), False
    while not terminal:
        actions = agent.act(next_state, deterministic=flat['deterministic'])  # test
        next_state, terminal, reward = env.execute(actions)
env.close()
agent.close()
```

So I could add a new function `evaluate()` or such? Thoughts? 

BTW for ^ to work, I needed to remove `Runner`'s call to `agent.close(), environment.close()` so I can keep it "dirty" while I evaluate, then manually close later (see https://github.com/lefnire/tensorforce/commit/ba3d45305141cd97d2c615de372c45588787ceea). I personally think it's a good idea to leave it up to the user anyway; want me to create a PR?

```
#Runner.py
def close():
    self.agent.close()
    self.environment.close()
````

lefnire:
Gonna close this for now until I come up with a better approach

AlexKuhnle:
Hmm, seems like the evaluation would probably happen locally, hence doesn't need to rely on the timestep/episode increment within the model -- which in fact probably should not happen, since evaluation should not influence exploration decay or so. A basic periodic evaluation capability like this could be integrated into the runner, so feel free to create a PR if you have a good solution. Any more complicated evaluation schemes for now should go into the benchmark project.

PR #265, opened 2017-12-13
summary_spec description

Mazecreator:
Add summary_spec description in Markdown format

### DIFF
--- a/SUMMARY_SPEC.md
+++ b/SUMMARY_SPEC.md
@@ -1,0 +1,107 @@
+TensorForce: Details for "summary_spec" agent parameters
+====================================================================
+
+[![Docs](https://readthedocs.org/projects/tensorforce/badge)](http://tensorforce.readthedocs.io/en/latest/)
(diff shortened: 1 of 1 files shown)


michaelschaarschmidt:
thanks! I might move it into a manuals folder where we can collect these

Mazecreator:
Sounds good, I just wanted to capture the information before it got forgotten.  I didn't know how to add it to the Docs.  Read the website but still couldn't figure out how to link it there... We should get one of these for Network_Spec too...

michaelschaarschmidt:
I think the website docs are auto-generated, which is @krfricke territory (ask in gitter)

krfricke:
We should move such file into the docs, either as a subchapter in the docs/agents_models.md file, or as a separate file. For networks it might make sense to add a neural_networks.md file or alike.
Anyway, such information belongs in the docs, otherwise the project root will be flooded. I'll move it to an appropriate location sometime this or next week.

PR #270, opened 2017-12-16
Fix categorical distribution

jjakimoto:
When  computing state_value,
`state_value = tf.reduce_logsumexp(input_tensor=logits, axis=-1)` is used. 
Does this have to be zero arrays for DQN?

### DIFF
--- a/tensorforce/core/distributions/categorical.py
+++ b/tensorforce/core/distributions/categorical.py
@@ -51,8 +51,7 @@ class Categorical(Distribution):
         shape = (-1,) + self.shape + (self.num_actions,)
         logits = tf.reshape(tensor=logits, shape=shape)
 
-        # !!!
-        state_value = tf.reduce_logsumexp(input_tensor=logits, axis=-1)
+        state_value = tf.zeros(tf.shape(logits[:-1]))
 
         # Softmax for corresponding probabilities
         probabilities = tf.nn.softmax(logits=logits, dim=-1)


michaelschaarschmidt:
Hm can you elaborate a little? This is not obviously true to me. For example, NAF needs state values:

https://github.com/reinforceio/tensorforce/blob/master/tensorforce/models/q_naf_model.py#L190

jjakimoto:
Hi @michaelschaarschmidt,
Thanks for your response.
I will look into that and figure out what would be the best.

AlexKuhnle:
Hi @jjakimoto,

There is a [dueling layer](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/core/networks/layer.py#L607) available, which is inspired by the dueling network architecture. It doesn't exactly correspond to what is described in the paper, though, which would be better implemented as a TensorForce [distribution](https://github.com/reinforceio/tensorforce/tree/master/tensorforce/core/distributions) probably. There was a discussion a while ago about this, with the aforementioned dueling layer as result, so maybe also worth getting in touch with who contributed it (@Mazecreator, I think). I remember that the current solution wasn't ideal, so improvements very welcome... :-)

Mazecreator:
That is correct, Dueling should be the output of the network to match the paper.  We might be able to accomplish this by changing the distribution layer a bit.  Maybe two options:
1) If the output of the network is the same as the number of actions required, don't apply a matrix from NxN, simply pass through the N without adding another matrix multiplication. (My vote)
2) Add a None distribution, much like the None Activation where the output of the network is directly passed (ultimately, deleting any distribution layer).

jjakimoto:
Hi @AlexKuhnle  and @Mazecreator ,
I think that my proposal is close to option 2 proposed by @Mazecreator. If the model has to use advantage and state functions, you add another distribution internally and use it inside the tf_q_value function. 

@Mazecreator, I do not understand your proposal 1 clearly. What does N stand for here?

Mazecreator:
@jjakimoto N is the number of actions.  So if the end of the "Network" has N outputs (namely from a dueling layer), then don't send them through a NxN matrix multiplication as it already has the proper size for the required actions.

jjakimoto:
@Mazecreator Understood.
In the original paper, after the embedding of lower layers, they split in two flows. Each flow has two layers, where middle layer could have arbitrary number of hidden units. 

In my opinion, we should have another network other than self.logits like self.state and then, we combine their outputs.

AlexKuhnle:
I actually wouldn't go for the 'implicit' option 1, because it sounds like a source for unintentional bugs, particularly if transferring a model to another environment (even if unlikely). A 'none' distribution might be something we want to support in general. However, a `categorical-dueling` distribution type using the dueling layer instead of a linear layer might be the best to make this feature generally available, or what do you think about that?

@jjakimoto, I would definitely prefer that the solution doesn't include two distributions per action, since that's not how distributions currently are supposed to work (although it makes sense from a practical standpoint). Since the motivation is to get both q values and state values, which are both provided by the distribution, it just appears to me to be a case for a new distribution type which does exactly that.

michaelschaarschmidt:
Should we close this PR since the original commit does not really relate to the topic of conversation any more?

michaelschaarschmidt:
Closing now, I suggest to re-open or create a different PR regarding the the duelling aspect, note that complex networks have been incorporated via @Mazecreator's PR

PR #280, opened 2017-12-22
Complex Network Implementation

Mazecreator:
ComplexNetwork basic implementation:

    network_spec_CL_selu_cp = [ 
        [
            dict(type='input', inputs=['state']),
            dict(type='dense', size=50, activation='selu', skip=False,l1_regularization=.00,l2_regularization=.0002),
            dict(type='dense', size=25, activation='selu', skip=False,l1_regularization=.00,l2_regularization=.0002),
            dict(type='dueling', size=3, activation='none', output=('expectation1','advantage1','mean_advantage2'),
                summary_labels=['activations','expectation','advantage','mean_advantage']),               
            dict(type='output', output='skip'),
            dict(type='dense', size=3, activation='selu', skip=False,l1_regularization=.00,l2_regularization=.0002),
            dict(type='input', inputs=['*','skip']),            
            dict(type='dense', size=3, activation='selu', skip=False,l1_regularization=.00,l2_regularization=.0002),
            dict(type='dueling', size=3, activation='none', output=('expectation2','advantage2','mean_advantage2'),
                summary_labels=['activations','expectation','advantage','mean_advantage']),
            dict(type='output', output='actions'),   
        ],
        [
            dict(type='input', inputs=['*','advantage1']), 
            #dict(type='input', inputs=['actions','advantage']),    # Same as above                    
            dict(type='dense', size=None, activation='selu', skip=True,l1_regularization=.00,l2_regularization=.0002),
            dict(type='dense', size=None, activation='selu', skip=True,l1_regularization=.00,l2_regularization=.0002),
            dict(type='dueling', size=3, activation='none', output=('expectation3','advantage3','mean_advantage3'),
                summary_labels=['activations','expectation','advantage','mean_advantage']),            
        ]    
    ] 

The ComplexNetwork is a List of List of Dict vs LayeredNetwork which is a List of Dict.
Output from the previous Dict() is available to the next Dict() regardless of network structure.

The Input layer will accept named inputs as a list where '*' or 'previous' means output from the previous dict.  This allows inline concatenation from earlier steps.  Outputs can be created inline as well to create skips, etc.  Some Layer types can support naming internal calculates for external use, such as "dueling".  The advantage will be useful in some later steps if this layer is used early in a network.

### DIFF
--- a/tensorforce/core/networks/__init__.py
+++ b/tensorforce/core/networks/__init__.py
@@ -16,6 +16,7 @@
 from tensorforce.core.networks.layer import Layer, Nonlinearity, Dropout, Flatten, Pool2d, Embedding, Linear, Dense, \
     Dueling, Conv1d, Conv2d, InternalLstm, Lstm
 from tensorforce.core.networks.network import Network, LayerBasedNetwork, LayeredNetwork
+from tensorforce.core.networks.complex_network import Input, Output
 
 
 layers = dict(
@@ -30,7 +31,9 @@ layers = dict(
     conv1d=Conv1d,
     conv2d=Conv2d,
     internal_lstm=InternalLstm,
-    lstm=Lstm
+    lstm=Lstm,
+    input=Input,
+    output=Output   
 )
 
 
@@ -41,12 +44,14 @@ __all__ = [
     'Dropout',
     'Flatten',
     'Pool2d',
+    'Embedding',
     'Linear',
     'Dense',
     'Dueling',
     'Conv1d',
     'Conv2d',
     'InternalLstm',
+    'Lstm',
     'Network',
     'LayerBasedNetwork',
     'LayeredNetwork'
--- a/tensorforce/core/networks/complex_network.py
+++ b/tensorforce/core/networks/complex_network.py
@@ -1,0 +1,187 @@
+# Copyright 2017 reinforce.io. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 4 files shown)


Mazecreator:
Here is a quick sample of DFP network structure.  The state dictionary would need 3 inputs; goal,measure,state.  Output is labeled as prediction.  Certainly there is more to the DFP on the cost function end as well as setup of the goal/measure, but the complex network structure will support work in that direction:

        network_spec_DFP = [ 
                        [
                            dict(type='input', inputs=['goal']),
                            dict(type='dense', size=6, activation='relu'),
                            dict(type='output', output='goalnet'),  
                        ],
                        [
                            dict(type='input', inputs=['measure']),
                            dict(type='dense', size=6, activation='relu'),
                            dict(type='output', output='measurenet'),  
                        ],
                        [
                            dict(type='input', inputs=['state']),
                            dict(type='dense', size=20, activation='relu'),
                            dict(type='dense', size=10, activation='relu'),
                            dict(type='output', output='statenet'),  
                        ],                
                        [
                            dict(type='input', inputs=['goalnet','measurenet','statenet']), 
                            dict(type='dense', size=40, activation='relu'),
                            dict(type='dense', size=20, activation='relu'),
                            dict(type='dueling', size=3, activation='none'),   
                            dict(type='output', output='prediction'),           
                        ]    
                            ]

AlexKuhnle:
Thanks, @Mazecreator, for the PR, very useful feature. There will likely be refinements over time, but you probably expect that. I think we might also decide to change some details, where at least I think things should be done more explicitly, like concatenating multiple inputs (however, we first need a concatenation layer), or using 'previous' as keyword instead of explicit references. The casting bit also needs some work, but you note that in the comments yourself.

Anyway, thanks very much, great work!

PR #281, opened 2017-12-22
ale.py crashes due to wrong variable referenced in episode_finished

cartershanklin:
Calculation of the step rate references an undefined variable in Runner. Maybe this variable used to exist but is not there in 0.3.3. It looks like r.timestep does the same thing.


michaelschaarschmidt:
Thanks - ALE is a contributed example and we do only do rudimentary API updates!

PR #283, opened 2017-12-22
ale.py --save never actually saves

cartershanklin:
Nothing ever calls agent.save_model

Borrowed code from the threaded runner that does the same thing.


michaelschaarschmidt:
Thanks! See comment in other PR

PR #286, opened 2017-12-26
Cleaup of `Agent` and its child classes

sven1977:
Moved some variables shared amongst all types of Agent (sub)classes into respective base class (e.g. `memory` goes into MemoryAgent instead of each single child of MemoryAgent). This saved a lot of redundant code and initialization procedures.
Made sure no variables defined in child classes are required by parent class (e.g. summary_spec).
Removed redundant docstring content, e.g. if variable is already defined by a parent class' c'tor and has the exact same functionality there.
Made sure the parent c'tor gets called as early on as possible within the child c'tor. This will prevent things like the above from happening.

I further updated the PR: I added a LearningAgent class which contains all the variables that should not(!) go into the non-learning classes (RandomAgent and ConstantAgent), e.g. discount, reward_preprocessing_spec, etc..). This again saved tons of redundant c'tor/init code in the child classes. As a last step, we could even think about creating Q-learning and Policy-gradient sub-classes (there are still some variables redundant related to Q-models and policy-grad. models).

Also gives an informative deprecation warning now when using DDQNAgent as it's redundant (use DQNAgent with double_q_model==True, instead).


michaelschaarschmidt:
Hey,
many thanks for the refactor, will have to look it over in more detail. 
First,  it seems like this causes a number of errors in the build:
https://travis-ci.org/reinforceio/tensorforce/jobs/321935104

You can generally run the tests (main build now passing, sometimes it takes a few attempts) to check for errors when you fix, or look at the travis log here

sven1977:
Ok, I'll take a look at the build problems.

AlexKuhnle:
Re conversation on Gitter: I would definitely not go down the route of introducing various intermediate agent classes, for two reasons:

(a) With the next memory update even `BatchAgent` and `MemoryAgent` will go, leaving the agent classes as mere wrappers which offer and constrain model creation in certain ways (for instance, DDQN vs DQN, essentially just one boolean flag pre-set in the former).

(b) One motivation (at least for me) behind these developments is to move away from the classic "Q model ~ replay memory" and "PG model ~ on-policy batches", and configuring agents/models that way would not support this move.

My views, obviously open to discussion, but I consider the agent classes as just transferring certain parameters and hardcoding others, and I hence don't mind the code replication in this one case.

Having said that, such classes might make sense to offer easy-to-use agent base classes of some sort, so some classes can make sense, but I wouldn't base them on model-specific aspects like Q vs PG agent.

AlexKuhnle:
An example of such a not-too-model-specific but reasonable base class could, for instance, be the `LearningAgent` you introduce (haven't looked into it yet, but I can generally see that this makes sense).

sven1977:
Great, thanks for the feedback.
I think I found a good way of avoiding too much replication and keeping the `Agent` classes - as you said - as mere default-parameters-setters. I just wanted to clean up and learn the code in the meantime. What you are saying about Q vs PolGrad makes  sense: I won't introduce any more intermediary classes.

Running it all through my local pytest and it's looking good. Sorry for the erroneous code yesterday. I'll probably do another (clean) push later today.

AlexKuhnle:
Okay with the PR, but three comments:
- I'm not a big fan of these `@staticmethod`'s, are they doing something other than moving the code?
- As mentioned in Gitter, the type system in documentation needs discussion before integrating it everywhere.
- I would encourage to discuss features like static methods, introduction of `LearningAgent`, etc before going ahead, as you did with attributes set to `None` in constructor (not sure whether discussion or change happened first, though ;-) ).

AlexKuhnle:
To clarify, in case this came across as 'negative': I think the `LearningAgent` and related changes are a good idea...  :+1:  :-)

sven1977:
Thanks for the feedback and the merge Alex! No offense taken 😄 
The reason why I'm finicky about these style issues is not because I invented these rules, but because I'm developing with PyCharm, which has very strict coding style checks in place (pep8 and many others, e.g. even if you leave two blank lines between functions instead of one, you get "reminded" by the IDE).
And I have come to really appreciate these rule-enforcements as they help reduce future errors (like the one we found today with the exploration never actually working in the naf tests).
I understand that at first it might seem unwieldy e.g. to define all class variables in __init__ even if only setting it to None at first. But it enforces you to have a fixed place for all class variables (nicely commented on what they are for and what they do) and makes it "harder" to unwillingly reuse the same one for different purposes. And all the other style rules have similar benefits.

sven1977:
- I will be more careful with future PRs (e.g. not being properly tested 😐 ).
- When should we discuss style guidelines (like @staticmethod, pep8, all variables defined in init, type hinting, etc..)?

AlexKuhnle:
I actually very appreciate that you are so strict about these style issues. My IDE also complains about every line which breaks PEP, and I'm quite annoyed if I find such cases.^^ The `None` thing is not part of PEP, so I don't get reminded, but I think it is a good idea to handle it as you did in this PR, so happy to add this to the (so far non-existent) style guidelines. Regarding features, I think it's just good to mention ideas before implementing them, and to some degree this should probably also happen the other way round, but let's leave the latter for beta version and later... :-)

I would be happy to get the guidelines started, finally, we could even do this via Skype. We can also put together a more general plan for the near future, or at least a list of feature ideas with some details.

sven1977:
Sounds great. I would be available via Skype or appear.in these days. Let me know.
Yes, sorry for my storminess with regards to pushing lots of changes at once without really waiting for discussion feedback. I'm just very excited about this project. 😃

AlexKuhnle:
Tomorrow could maybe work, but will let you know.

PR #289, opened 2017-12-29
Fixed problems with exploration.

sven1977:
- Model.action_exploration was not defined -> changed it to fn_action_exploration
- This lead to a couple of other hidden problems in the Exploration classes where sometimes, an int was assigned to a shape-parameters in a tf function, which lead to runtime errors.
- Changed num_actions in tf_explore to action_shape to account for that.
- also had to add a few casts (int and float) to make sure tf types are not conflicting (sometimes, when a default value for a function's parameter is 0.0, it changes it to int=0)
- pytests are all "pass"

### DIFF
--- a/tensorforce/core/explorations/constant.py
+++ b/tensorforce/core/explorations/constant.py
@@ -25,5 +25,5 @@ class Constant(Exploration):
         self.constant = constant
         super(Constant, self).__init__(scope=scope, summary_labels=summary_labels)
 
-    def tf_explore(self, episode, timestep, num_actions):
+    def tf_explore(self, episode, timestep, action_shape):
         return self.constant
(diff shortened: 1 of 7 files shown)


michaelschaarschmidt:
Hey, many thanks!
In OU exploration, you are casting with float(mu).

Would it not be cleaner to do util.tf_dtype('float') so we ensure same precision wherever we cast?

michaelschaarschmidt:
Ok upon further consideration that cast is probably fine - merging

sven1977:
These casts are definitely debatable. Not a big fan of them either, but tf was crashing because both tf types for the "Greater-than" op have to be the same (int with float does not work). Strange that python treats a 0.0 as 0 when given as a default value in the parameters of a func (mu=0.0).

PR #292, opened 2018-01-02
added n-step cumulative discounted reward to Model and completed commenting on model.py and runner.py class(es)

sven1977:


michaelschaarschmidt:
Hey, many thanks for this, will take a few days to get to potentially. Just a first comment:

We sqash-merged one of the first PRs you made because it contained many small commits, which means the commit history is now different in the master than in your branch. Your PRs seem to include all prior commits on your branch so I think this can cause problems merging going forward. I think it would be ideal to have every PR from a 'clean' rebase of the master and only include the commits relevant to the PR. Also makes it much easier to see just the changes made in this PR.

sven1977:
Agreed. Sorry, I saw this as well now. I did do a full rebase of my repo against the upstream master before I filed this PR. But something else may be wrong. I'll take another look and resubmit.

PR #294, opened 2018-01-02
First look at optional logits return

Mazecreator:
This PR will return logits via call to agent act() if include_logits=True.

Not sure if there are other parts of the code that need updating.  I could only find the "distribution_model" to be responsible for logits.

### DIFF
--- a/tensorforce/agents/agent.py
+++ b/tensorforce/agents/agent.py
@@ -154,7 +154,7 @@ class Agent(object):
         # for preprocessing in self.preprocessing.values():
         #     preprocessing.reset()
 
-    def act(self, states, deterministic=False):
+    def act(self, states, deterministic=False, include_logits=False):
         """
         Return action(s) for given state(s). States preprocessing and exploration are applied if  
         configured accordingly.
@@ -162,9 +162,10 @@ class Agent(object):
         Args:
             states: One state (usually a value tuple) or dict of states if multiple states are expected.
             deterministic: If true, no exploration and sampling is applied.
+            include_logits: If true, logits are returned with action
         Returns:
             Scalar value of the action or dict of multiple actions the agent wants to execute.
-
+            (logits) Optional second value of logits returned if include_logits = True
         """
         self.current_internals = self.next_internals
 
@@ -173,17 +174,31 @@ class Agent(object):
         else:
             self.current_states = {name: np.asarray(state) for name, state in states.items()}
 
-        # Retrieve action
-        self.current_actions, self.next_internals, self.timestep = self.model.act(
-            states=self.current_states,
-            internals=self.current_internals,
-            deterministic=deterministic
-        )
+        if include_logits:
+            # Retrieve action
+            self.current_actions, self.next_internals, self.timestep, self.logits = self.model.act(
+                states=self.current_states,
+                internals=self.current_internals,
+                deterministic=deterministic,
+                include_logits=True
+            )
 
-        if self.unique_action:
-            return self.current_actions['action']
+            if self.unique_action:
+                return self.current_actions['action'], self.logits
+            else:
+                return self.current_actions, self.logits            
         else:
-            return self.current_actions
+            # Retrieve action
+            self.current_actions, self.next_internals, self.timestep = self.model.act(
+                states=self.current_states,
+                internals=self.current_internals,
+                deterministic=deterministic
+            )
+
+            if self.unique_action:
+                return self.current_actions['action']
+            else:
+                return self.current_actions
 
     def observe(self, terminal, reward):
         """
--- a/tensorforce/agents/agent.py
+++ b/tensorforce/agents/agent.py
@@ -154,7 +154,7 @@ class Agent(object):
         # for preprocessing in self.preprocessing.values():
         #     preprocessing.reset()
 
-    def act(self, states, deterministic=False, include_logits=False):
+    def act(self, states, deterministic=False, fetch_tensors=None):
         """
         Return action(s) for given state(s). States preprocessing and exploration are applied if  
         configured accordingly.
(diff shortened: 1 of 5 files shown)


michaelschaarschmidt:
Thanks for putting up PR - heads up: will take till weekend for next session for us to be able to discuss it, because we need to decide whether we want to change the act API or find another way.

Mazecreator:
No problem, this should be backward compatible with all existing software and I made sure I wasn't moving data unless it was asked for so it shouldn't impact anyone's performance either.

If there is a better place or refactoring for this data even better.  I can use this code for now for debugging.

ogutty:
Thanks for the code, I was able to get my hands on the logits from within runner.py.
You may want to update runner.py as well, specifically line 96 needs to ask for the logits, and accept them if & when they are returned:

michaelschaarschmidt:
Ok so a first higher level comment:

Expanding the act() api should be done via carefully, and I think creating additional keywords per extra-requested value is not ideal.

 Something I would prefer as a more generic situation is to have an optional argument like ```meta_data_``` which is a dict, and in that dict one could then pass keys and additional options for the meta_data., e.g. a key 'logits'. 

The model would then optionally return a dict of meta data with metadata field names and values. Does that make sense?

Mazecreator:
@ogutty, I needed to customize my application to use the Logits for export.  What do you want to do with the logits from the runner class?  I would not want to break what appears to be a helper function, requesting the logits could increase data IO and slow things down a little so I didn't want to work on other loops which might get slowed as well.

@michaelschaarschmidt I can't say I am a big fan of passing data through arguments, but if that is what we need to do I guess it can be done.  I used the separate keywords to request the data as when I was adding more summary details to the optimizer I was asked to not use kwargs as to force a hard failure rather than a possible hidden failure (due to spelling, etc).

If we are reconsidering this for the architecture I can certainly change it.  Maybe the contribution.md file can be updated to include a bit of a discussion along this line as well?

michaelschaarschmidt:
Why would we pass data through arguments? We would pass an argument specifying which meta data fields we are interested in (of which logits are just one and we could think of others, such as raw network outputs), and then we return all required meta data fields in a result dict

Mazecreator:
Okay, I miss understood using the arguments comment.  So pass a kwarg or something that can specify what is needed, and return a second tuple() or something from the .act()?

    action, logits = env.act(....., meta_data = {'logits': True} ) 

Something like this?  With the possibility of extending:
     action, data= env.act(....., meta_data = {'logits': True, 'myfavoratedata': True} )
     # data will contain a tuple() of logits, myfavoratedata  (in this case the order isn't clear)

Not sure how to resolve sequence of data returned in tuple() do we have a fixed sequence?

Mazecreator:
Sorry, just saw you wanted a dict() returned... That makes better sense...

If this is agreed I can get to work on it?  Would only one "dict()" be returned if additional data is requested?  If none, just the pure "action", so we would bundle { 'action': action, 'logits': logits, ...}

michaelschaarschmidt:
Yes, meta_data = {'logits': True}. I initially thought there might be potential additional parameters on the meta data, but if that's not the case, we could also simply pass a list of meta data labels, meta_data = ['logits'], but then I would have returned a dict with name of the requested meta data field and its value, i.e. data = { 'logits' : values}

Mazecreator:
I think the dict() in might be more flexible as you would imaging someone wanting a "slice" or something in the future.  Or maybe even a named variable from the Complex Network?

I am struggling from a cold so bear with me, also bitter cold on the East Coast US.... SOrry to be a pain today...

Just on the last point?  we return a "action" if no meta_data requested, otherwise bundle the returned values?

ogutty:
@Mazecreator , you're probably right regarding (not) integrating into runner.py (again, I'm still feeling my way into the project).  I was using the runner as a default "gateway", and once my application matures I'll probably bypass it as well.

Mazecreator:
New API still backward compatible with old style if fetch_tensors=None.  Named output from the complex network can also be listed to fetch.

    action, fetches = agent.act(state,deterministic=True, fetch_tensors=['logits'])
    logits = fetches['logits']

make named_tensors available to Network class and now set_named_tensor() used to store logits and other model.py data.

michaelschaarschmidt:
Seems like there are some conflicts now that would need resolving before we can go ahead?

Mazecreator:
Should be set now.

AlexKuhnle:
Hi @Mazecreator, I would encourage resembling the e.g. `get_summaries()` interface more, so `get_tensor(self, name)` for `Distribution` as well as for `Network`, `Layer`, etc. `Model.get_tensor` then internally calls e.g. `self.network.get_tensor` which itself internally calls `layer.get_tensor` for all its layers. `get_list_of_named_tensor` seems unnecessary apart from the error message, and `set_named_tensor` should not be necessary -- every class can internally set e.g. `self.tensors[name] = logits`, constructor initializes `self.tensors = dict()`. Checks like `if self.network is not None and fetch_tensors is not None:` in `Model` will definitely throw exceptions for models which don't use networks (e.g. `RandomModel`), so should not exist in that way.

Does this sound sensible, and would you move things accordingly? (Otherwise I can do that at some point...)

Mazecreator:
@AlexKuhnle I am not sure I fully understand your message here.  I will take a closer look this weekend or early next week.  I will likely contact you on gitter for more advice and clarification.  I think I understand portions, but there are a few cases that may not be clear from what I remember due to class definitions and inheritance.  Doesn't sound unreasonable, but I want to make sure I follow your lead to make sure I only spin this one more time.

As for the get_list_of_named_tensor(), that is for the error.  The issue is there are a few cases when the user might not be able to guess the tensor name (like logits for example with multiple distributions) as they will be changed from "logits" to "action_logits", etc.  This is to help speed up the processes of building the network.  I think this form of help is very useful, so I am not certain what to do, maybe that one is called directly or I might be able to move it if we have self.tensors[] available in model.

AlexKuhnle:
Yes, the second point about the named tensor list makes sense, so we could leave it as it is. What I mean re the first point is that registering tensors coming from a distribution in the network [here](https://github.com/reinforceio/tensorforce/pull/294/commits/7ee8b5c8966f0573bfde0c5e1b4fd804dc1b83e7#diff-60e3618cbf79c10a2e295d2b61c6a4b2R156) seems weird, and we could instead let every class provide the named tensors coming from it, as we do for variables or summaries. Although, looking at it again, I'm not 100% sure what would be the best way to do so. If you're not sure, we can also merge and later on modify it...

michaelschaarschmidt:
Any updates on this?

Mazecreator:
@AlexKuhnle was going to test this branch and then merge most likely after the memory PR goes through.  We were going to make changes at that time to put in new functionality.

AlexKuhnle:
Merged.

PR #296, opened 2018-01-04
Add swish non-linear activation function

Mazecreator:
Added swish as described here:
https://arxiv.org/abs/1710.05941

This paper describes a "Beta" parameter as well which technically could be applied to any non-linearity. 
 This may be something we want to consider enabling, doesn't make sense in some activations but others it does.  I will open another discussion for a "learnable Beta" value.


AlexKuhnle:
Is it okay if we leave this open for you to add handling the beta parameter as proposed in Gitter?

Mazecreator:
This is obsolete, see new PR.

PR #298, opened 2018-01-05
Learnable Beta plus a few more TF activation functions

Mazecreator:
Added swish activation as well as learnable Beta.  Beta parameter can also be added as a constant through the new "activation_kwargs".  A few new TF activations added:
lrelu/leaky_relu: kwargs 'alpha' can be set for leaky slope
crelu
softsign

kwargs supports the follow:
                        {'beta': float|int|'learn'} # for beta factor on input
                        {'alpha': float|int}   # For leaky Relu
                        {'max': float|int}   # Clip maximum input to activation at "max"
                        {'min': float|int}   # Clip minimum input to activation at "min"

min/max have been added to clip the input values.  So for example, if you want a tf.nn.relu6, use a relu with a {'max': 6.0}.

New summary_labels 'beta' will show a scalar graph in tensorflow of the learned beta parameter.

### DIFF
--- a/tensorforce/core/networks/layer.py
+++ b/tensorforce/core/networks/layer.py
@@ -157,28 +157,31 @@ class Nonlinearity(Layer):
     Non-linearity layer applying a non-linear transformation.
     """
 
-    def __init__(self, name='relu', scope='nonlinearity', kwargs=None, summary_labels=()):
+    def __init__(self, kwargs='relu', scope='nonlinearity', summary_labels=()):
         """
         Non-linearity layer.
 
         Args:
-            name: Non-linearity name, one of 'elu', 'relu', 'selu', 'sigmoid', 'softmax', 'softplus', 'tanh' or 'none'.
-            kwargs: Optional dictionary of args for activation; Currently supports;
+            kwargs: Non-linearity name, one of 'elu', 'relu', 'selu', 'sigmoid', 'softmax', 'softplus', 'tanh' or 'none'.
+                    Optional dictionary of args for activation, must include {'name':'xxx'} Currently supports;
                         {'beta': float|int|'learn'} # for beta factor on input
                         {'alpha': float|int}   # For leaky Relu
                         {'max': float|int}   # Clip maximum input to activation at "max"
                         {'min': float|int}   # Clip minimum input to activation at "min"
             summary_labels: Requested summary labels for tensorboard export
         """
-        self.name           = name
+        if type(kwargs) is str:
+            self.name       = kwargs
+            #kwargs          = dict()
+        else:
+            self.name       = kwargs['name']
+
         self.alpha          = None
         self.max            = None
         self.min            = None        
         self.beta_learn     = False
         super(Nonlinearity, self).__init__(scope=scope, summary_labels=summary_labels)
 
-        if kwargs is None:
-            kwargs = dict()
         if 'max' in kwargs:
             if type(kwargs['max']) is int or type(kwargs['max']) is float:
                 self.max = float(kwargs['max'])       
@@ -587,7 +590,6 @@ class Dense(Layer):
         size=None,
         bias=True,
         activation='tanh',
-        activation_kwargs=None,
         l2_regularization=0.0,
         l1_regularization=0.0,
         skip=False,
@@ -600,8 +602,7 @@ class Dense(Layer):
         Args:
             size: Layer size, if None than input size matches the output size of the layer
             bias: If true, bias is added.
-            activation: Type of nonlinearity.
-            activation_kwargs: Arguments passed to nonlinearity
+            activation: Type of nonlinearity, or dict with name & arguments
             l2_regularization: L2 regularization weight.
             l1_regularization: L1 regularization weight.
             skip: Add skip connection like ResNet (https://arxiv.org/pdf/1512.03385.pdf),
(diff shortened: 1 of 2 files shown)


Mazecreator:
Reworked activation definition:
dict(type='dense', size=15, activation={'name':'swish', 'beta': 1.9} )
dict(type='dense', size=15, activation={'name':'swish' ,'beta': 'learn'} )
dict(type='dense', size=10, activation={'name':'leaky_relu', 'max': 10.0, 'alpha': 0.31}
dict(type='dense', size=10, activation={'name':'lrelu', 'max': 10.0, 'min':-10.0} )

AlexKuhnle:
Hi @Mazecreator,

First, two details: We always want operator spacing (for `*` in your code) and named arguments for Tensorflow functions (`tf.minimum` etc in your code). Second, I would definitely not use `kwargs` as parameter to `Nonlinearity`, but instead spell all the arguments out and pass `activation` in e.g. `Dense` via `**activation`. Third, I'm thinking whether these `min` and `max` values should really be a separate layer `clip_value` or so. Fourth, is the multiplication with `self.beta` really what you want here: `x = tf.minimum((self.beta*x), self.max)`?

Mazecreator:
@AlexKuhnle I will try to get to this over the weekend as I have a few other things to accomplish ahead of this one.  I think I understand what you are saying here so "self.beta * x", and x = tf.minimum(x=(self.beta * x), y=self.max).

I think the min/max should also be a layer but also belong as part of the activation.  The inspiration for this addition is "relu6".  I didn't want to add an arbitrary upper clipping of 6, so to make this you would use a relu with a max of 6 (you can change it to 7 or whatever now).  The min is there in case there is desire to explore in this area too.  I think adding another layer feels like another layer not an activation function.  But I do believe there are times for clarity, a new layer would be a better approach.  It certainly would add flexibility.

As for the forth point... I am open to suggestions... My thought was if we were to put in a "relu6" for example but beta is 2.0, then setting a max of 6 would clip a value of 3.0 * 2.  If the beta is not there, then the clipping would actually be at 12.0.  I am flexible on this, that was my thinking, but not sure it is a perfect reasoning.  I was simply considering the input as (X * BETA), this might not be appropriate for more complex activation functions such as elu(), etc...  Here, a separate layer would not be able to handle this situation if beta is trainable.

AlexKuhnle:
Okay, we can definitely leave `min` and `max` as they are for now (could still change it later, if we wanted). I'm also not sure re the last point, I just thought it's not what I would have expected as behavior to multiply twice with `beta` if `min` or `max` are set, and I think it is indeed not quite what one wants, right? Other than that, happy to merge after the other 1-2 minor corrections, thanks!

Mazecreator:
@AlexKuhnle I was hoping to avoid a lot of overhead in layers without a Nonlinearity.  By passing hard parameters in the INIT it is going to be tricky to handle the backward compatibility for the "type='relu'" for example.  I need to check in each impacted layer if the type is a dict or string.  And figure if it is a string, make it an activation['name'] value to pass into the Nonlinearity.  Is this your expectation or show I maybe create a UTIL which does just this function?

AlexKuhnle:
Hmm, let me check what exactly it would look like, and probably we can just merge the PR and later on improve this bit, if we want. What I mean is a general mechanism as we have for specs generally, where it's possible to say `optimizer='adam'`  or `optimizer=dict(type='adam', ...)`. It seems a good idea for layers to allow shallow and detailed configurability for sub-layers of more complex layers without argument overhead. If it's a good idea, no need to worry too much about backward compatibility (we're still early alpha :-).

Mazecreator:
I do like the backward compatibility so we don't break all older networks.  I think we can do it, but to eliminate the redundant code I will create a utility to make the conversion "prepare_activation()".  This will allow the activation to be handled the same way as the optimizer.  A more generic version of that utility might be handy to cover other cases if there are any in the future.  But for now, I will make it specifically for the Activation.

Mazecreator:
Made the conversions for the hard params and pep8.  Added a util to prepare kwargs to eliminate redundant code but support passing string or dict into activation parameter.

michaelschaarschmidt:
Hey, looks like there are some build problems with this:
'    from tensorforce.core.networks.layer import Layer, Nonlinearity, Dropout, Flatten, Pool2d, Embedding, Linear, Dense, \
E     File "/tensorforce/tensorforce/core/networks/layer.py", line 624
E       self.nonlinearity = Nonlinearity(**util.prepare_kwargs(activation), summary_labels=summary_labels)
E                                                                         ^
E   SyntaxError: invalid syntax'

Mazecreator:
@michaelschaarschmidt any idea what the issue is here?  I can't tell which Python version threw this error.  The code is fine on 3.5 as this is simply a (**activation) where activation is a function that returns a dictionary.  I can try moving the ** to the end of the function call?  Not sure if this is a 2.7 error or not?

michaelschaarschmidt:
Don't really know, sorry, are you using PyCharm or similar? That should give you the relevant compatibility warnings and hints

Mazecreator:
I don't have PyCharm... 

At this point if you think this is going to break other Python versions, I would request if you can delete/reject the https://github.com/reinforceio/tensorforce/pull/298/commits/07f2ef70de1c30c687cdfb7951ff197daceb5e96 commit and commit it as originally proposed.  We can figure out the kwargs later.  I don't have a way to debug this issue as it doesn't fail on my system.  We could also ask if there are others out there that might be able to test 2.7 and 3.6 with this commit and give feedback of what may be the problem.  Ultimately, this is simply a **kwargs argument only with a function rather than kwargs.

AlexKuhnle:
I think swapping the arguments is what needs to be done. If I remember correctly, `**` arguments should always be after specified named arguments.

michaelschaarschmidt:
Just wanted to add that there is a pycharm community edition, and it is really worth it for debugging and the hinting

Mazecreator:
I will look into a pycham, I use Visual Studio Code now.  I will swap the two params and give that a shot.

lefnire:
@Mazecreator is this by default the same as before, if we don't change/use any of the new args? (ie,  will pulling this commit not change anything out of the box?)

Mazecreator:
@lefnire this commit should be total backward compatible and not impact any prior networks/results.

The recommended use for Swish is with no arguments or if you dare with the "learn" argument for beta.

PR #305, opened 2018-01-18
Add xavier_initializer and Try to add episode reward summary

yellowbee686:
Add xavier_initializer in Linear Layer and a parameter linear_weights in Dense, which is used for pass weight_init from Dense to Linear.

And I try to add episode summary, but it'll cause an error when first q_model.update(), I don't know what that error means, and I'm not sure whether I should use tf.make_template for calculating episode reward,
 I will be very grateful if anyone helps me.

error as follows:
InvalidArgumentError : Retval[0] does not have value
trace=['  File "H:\\LX6\\bin\\Debug\\lx6_init.py", line 50, in init\n    runner.run()\n', '  File "d:\\tensorforce\\tensorforce\\execution\\runner.py", line 127, in run\n    self.agent.observe(terminal=terminal, reward=reward)\n', '  File "d:\\tensorforce\\tensorforce\\agents\\memory_agent.py", line 145, in observe\n    return_loss_per_instance=True\n', '  File "d:\\tensorforce\\tensorforce\\models\\q_model.py", line 320, in update\n    fetched = self.monitored_session.run(fetches=fetches, feed_dict=feed_dict)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\training\\monitored_session.py", line 521, in run\n    run_metadata=run_metadata)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\training\\monitored_session.py", line 967, in run\n    raise six.reraise(*original_exc_info)\n', '  File "D:\\Python36\\lib\\site-packages\\six.py", line 693, in reraise\n    raise value\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\training\\monitored_session.py", line 952, in run\n    return self._sess.run(*args, **kwargs)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\training\\monitored_session.py", line 1024, in run\n    run_metadata=run_metadata)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\training\\monitored_session.py", line 827, in run\n    return self._sess.run(*args, **kwargs)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\client\\session.py", line 889, in run\n    run_metadata_ptr)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\client\\session.py", line 1120, in _run\n    feed_dict_tensor, options, run_metadata)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\client\\session.py", line 1317, in _do_run\n    options, run_metadata)\n', '  File "D:\\Python36\\lib\\site-packages\\tensorflow\\python\\client\\session.py", line 1336, in _do_call\n    raise type(e)(node_def, op, message)\n']

### DIFF
--- a/tensorforce/models/model.py
+++ b/tensorforce/models/model.py
@@ -176,6 +176,14 @@ class Model(object):
         # Inputs and internals
         # Current episode number as int Tensor
         self.episode = None
+        # episode reward as float Tensor
+        self.episode_reward = None
+        # TensorFlow op incrementing `self.episode_reward`
+        self.increment_episode_reward = None
+        # TensorFlow op reset `self.episode_reward`
+        self.reset_episode_reward = None
+        # Template function calculating episode rewards
+        self.fn_episode_reward = None
         # TensorFlow op incrementing `self.episode` depending on True is-terminal signals
         self.increment_episode = None
         # Int Tensor representing the total timestep (over all episodes)
@@ -322,6 +330,20 @@ class Model(object):
                 assert len(collection) == 1
                 self.episode = collection[0]
 
+            # episode_reward
+            collection = self.graph.get_collection(name='episode_reward')
+            if len(collection) == 0:
+                self.episode_reward = tf.Variable(
+                    name='episode_reward',
+                    dtype=util.tf_dtype('float'),
+                    trainable=False,
+                    initial_value=0.
+                )
+                self.graph.add_to_collection(name='episode_reward', value=self.episode_reward)
+            else:
+                assert len(collection) == 1
+                self.episode_reward = collection[0]
+
             # Timestep
             collection = self.graph.get_collection(name='timestep')
             if len(collection) == 0:
@@ -425,9 +447,10 @@ class Model(object):
 
             # Global trainables (from global_model)
             global_variables = self.global_model.get_variables(include_non_trainable=True) +\
-                               [self.episode, self.timestep]
+                               [self.episode, self.timestep, self.episode_reward]
             # Local counterparts
-            local_variables = self.get_variables(include_non_trainable=True) + [self.episode, self.timestep]
+            local_variables = self.get_variables(include_non_trainable=True) +\
+                              [self.episode, self.timestep, self.episode_reward]
             init_op = tf.variables_initializer(var_list=global_variables)
             ready_op = tf.report_uninitialized_variables(var_list=(global_variables + local_variables))
             ready_for_local_init_op = tf.report_uninitialized_variables(var_list=global_variables)
@@ -437,7 +460,8 @@ class Model(object):
                                        for local_var, global_var in zip(local_variables, global_variables)))
         # Local variables initialize operations (no global_model).
         else:
-            global_variables = self.get_variables(include_non_trainable=True) + [self.episode, self.timestep]
+            global_variables = self.get_variables(include_non_trainable=True) +\
+                               [self.episode, self.timestep, self.episode_reward]
             init_op = tf.variables_initializer(var_list=global_variables)
             ready_op = tf.report_uninitialized_variables(var_list=global_variables)
             # TODO(Michael) TensorFlow template hotfix following 1.5.0rc0
@@ -755,6 +779,11 @@ class Model(object):
             func_=self.tf_preprocess_reward,
             custom_getter_=custom_getter
         )
+        # self.fn_episode_reward = tf.make_template(
+        #     name_=(self.scope + '/episode_reward'),
+        #     func_=self.tf_episode_reward,
+        #     custom_getter_=custom_getter
+        # )
 
         self.summary_configuration_op = None
         if self.summary_spec and 'meta_param_recorder_class' in self.summary_spec:
@@ -832,6 +861,24 @@ class Model(object):
             reward = self.reward_preprocessing.process(tensor=reward)
 
         return reward
+
+    def tf_episode_reward(self, terminal, reward):
+        reach_terminal = terminal[-1]
+        reward_sum = tf.reduce_sum(input_tensor=reward)
+        episode_reward = self.episode_reward.assign_add(delta=reward_sum)
+        tf.cond(pred=reach_terminal, true_fn=(lambda: self.summary_episode_reward(episode_reward=episode_reward)),
+                false_fn=(lambda: episode_reward))
+        return episode_reward
+
+    def summary_episode_reward(self, episode_reward):
+        if 'episode-reward' in self.summary_labels:
+            summary = tf.summary.scalar(name='episode-reward', tensor=episode_reward)
+            self.summaries.append(summary)
+        return episode_reward
+
+    def tf_reset_episode_reward(self):
+        assign_op = self.episode_reward.assign(0.)
+        return assign_op
 
     # TODO: this could be a utility helper function if we remove self.discount and only allow external discount-value input
     def tf_discounted_cumulative_reward(self, terminal, reward, discount=None, final_reward=0.0, horizon=0):
@@ -1163,6 +1210,11 @@ class Model(object):
         with tf.control_dependencies(control_inputs=(increment_episode,)):
             self.increment_episode = self.episode + 0
         # TODO: add up rewards per episode and add summary_label 'episode-reward'
+        self.increment_episode_reward = self.tf_episode_reward(
+            terminal=terminal,
+            reward=reward
+        )
+        self.reset_episode_reward = self.tf_reset_episode_reward()
 
         # Tensor(s) fetched for model.update()
         self.optimization = self.fn_optimization(
@@ -1238,7 +1290,7 @@ class Model(object):
                 Current episode, timestep counter and the shallow-copied list of internal state initialization Tensors.
         """
         # TODO preprocessing reset call moved from agent
-        episode, timestep = self.monitored_session.run(fetches=(self.episode, self.timestep))
+        episode, timestep, _ = self.monitored_session.run(fetches=(self.episode, self.timestep, self.reset_episode_reward))
         return episode, timestep, list(self.internals_init)
 
     def act(self, states, internals, deterministic=False, fetch_tensors=None):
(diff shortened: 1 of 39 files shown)


Mazecreator:
@yellowbee686 
I have a PR in for leaky_relu() here https://github.com/reinforceio/tensorforce/pull/298
That portion of this PR is going to cause a conflict depending upon the order by which they are accepted.  The PR above will add Alpha for leaky_relu.  With some luck we can get #298 gets committed shortly.

yellowbee686:
@Mazecreator Ok, I'll wait for it, then change my PR, Thanks!

tensorforce/models/model.py
```diff
@@ -69,21 +69,21 @@ class Model(object):
...
-        saver_spec=None,
-        summary_spec=None,
-        distributed_spec=None,
-        optimizer=None,
-        discount=0.0,
-        variable_noise=None,
-        states_preprocessing_spec=None,
-        explorations_spec=None,
-        reward_preprocessing_spec=None
+            self,
```
michaelschaarschmidt: Hey is there a reason for these extra tabs?

michaelschaarschmidt:
Hey, I have had a look now and it seems like you changed quite a lot of formatting, especially adding/removing tabs in various places, e.g. you added tabs in the model constructor (also see some correct fixes to previously wrong formats). 

Could you change these tabs back so they conform to the 1 indent per level again?

AlexKuhnle:
You also changed or introduced some indents which we handle differently, e.g.:
```
            local_variables = self.get_variables(include_non_trainable=True) + \
                              [self.episode, self.timestep, self.episode_reward]
```
or: 
```
             feed_dict.update(
                {internal_input: (internals[n],)
                 for n, internal_input in enumerate(self.internals_input)}
             )
```

We prefer purely tab-based indents (instead of previous-line-alignments), so the following:
```
            local_variables = self.get_variables(include_non_trainable=True) + \
                [self.episode, self.timestep, self.episode_reward]
```
and
```
             feed_dict.update({
                internal_input: (internals[n],)
                for n, internal_input in enumerate(self.internals_input)
            })
```
If you could change that, that would be great, but also just to make you aware of our ['guidelines'](https://github.com/reinforceio/tensorforce/blob/master/CONTRIBUTING.md) (which definitely need to be extended...).

yellowbee686:
@michaelschaarschmidt @AlexKuhnle Thanks for replying, I'll change my code as the guidelines, and then check if the error still appears.

yellowbee686:
@michaelschaarschmidt I've change my code formatting, I think it's because I change the code on main branch and paste it to my own branch to commit, then my PyCharm do some auto formatting, which has some error.
@AlexKuhnle the code before my change is just like the fisrt format, but I change them as your advice.

Any way, do you have any idea about the error which my paste in front

tensorforce/models/model.py
```diff
@@ -830,6 +859,26 @@ def tf_preprocess_reward(self, states, internals, terminal, reward):
...
         return reward
 
+    def tf_episode_reward(self, terminal, reward):
+        reach_terminal = terminal[-1]
+        reward_sum = tf.reduce_sum(input_tensor=reward)
+        episode_reward = self.episode_reward.assign_add(delta=reward_sum)
+        return tf.cond(
+            pred=reach_terminal,
+            true_fn=(lambda: self.summary_episode_reward(episode_reward=episode_reward)),
+            false_fn=(lambda: episode_reward)
```
yellowbee686: I've debug that the error InvalidArgumentError : Retval[0] does not have value is because the false_fn of tf.cond, but I can't figure out what's wrong with it, do you have any idea about it? @michaelschaarschmidt  @AlexKuhnle   Thanks!

Mazecreator: @yellowbee686 if I recall from the specs of the TF Assign ops, "episode_reward" as you have it above is the Op to do the Assign_add and not the actual resulting value of the addition.  So false_fn is only assigning the value and returning that tf.assign_add() op.  I think you need to use a control flow to do what you are tryinig:

    with tf.control_dependencies([episode_reward]):
        return self.episode_reward

episode_reward is the assign_add Op and not the result.  The result will be in self.episode_reward according to your code.

yellowbee686: @Mazecreator Thank you very much, I'll try it!

yellowbee686: def tf_episode_reward(self, terminal, reward):
        reach_terminal = terminal[-1]
        reward_sum = tf.reduce_sum(input_tensor=reward)
        episode_reward = self.episode_reward.assign_add(delta=reward_sum)
        return tf.cond(
            pred=reach_terminal,
            true_fn=(lambda: self.summary_episode_reward(episode_reward=episode_reward)),
            false_fn=(lambda: self.assign_episode_reward(episode_reward=episode_reward))
        )

    def assign_episode_reward(self, episode_reward):
        with tf.control_dependencies(control_inputs=(episode_reward,)):
            return self.episode_reward

    def summary_episode_reward(self, episode_reward):
        if 'episode-reward' in self.summary_labels:
            summary = tf.summary.scalar(name='episode-reward', tensor=episode_reward)
            self.summaries.append(summary)
        return episode_reward

@Mazecreator I've tried as your advice and the error still occurs, here I try to return the assign_op and in observe I'll do the assign op in model.observe by 
episode, _ = self.monitored_session.run(fetches=(self.increment_episode, self.increment_episode_reward), feed_dict=feed_dict)

It's quite weird that if I change the code to
return tf.cond(
            pred=reach_terminal,
            true_fn=(lambda: self.summary_episode_reward(episode_reward=episode_reward)),
            false_fn=(lambda: self.summary_episode_reward(episode_reward=episode_reward))
        )
the error is the same

but if I change the code to:
return tf.cond(
            pred=tf.constant(value=True, dtype=tf.bool),
            true_fn=(lambda: self.summary_episode_reward(episode_reward=episode_reward)),
            false_fn=(lambda: self.assign_episode_reward(episode_reward=episode_reward))
        )
it works.

Another weird thing is I run the op in model.observe, but the error comes the first time do the q_model.update then I use dqn

Mazecreator: I think I see what you are trying to do but I am not sure it works the way you are trying to implement the change.  The summary ops are handled separately from the computational graph.  They will be fired I think after each episode is complete, I don't think there is a way to not pass the "episode_reward" to the summary when there is a non-terminal event.

yellowbee686: @Mazecreator I try to do the thing is to add up episode reward when non-terminal observation, and when terminal observation, summary it, so I try to return the assign_add op to session.run in model.observe, I'm a starter on tensorflow, so I'm not sure my code is right,  in the false_fn of tf.cond I only want to pass the assign_add op to model.observe, you can review the code in the PR, and thanks very much for your help

Mazecreator: The summary system is build into TF with a form of a callback.   Once you append it to the summaries() will be updated each episode regardless  of you tf.cond.  You  have formed an  operation in the TF graph and it will continue to be executed by the core TF RUN() callback.  The python code gets called once during the creation of the graph and builds the graph.  I don't think  this code is working the way you expect and I don't know of a way to fix it (so it only records reward upon terminal).  It just doesn't feel like it will fit into the structure right now with the summaries.

@AlexKuhnle  or @michaelschaarschmidt  might be able to chime in, but I don't see a way to only record the "episode_reward" upon terminal.  At best, you might be able to record 0 when not terminal and then episode_reward when terminal.  Not sure NA will work in TensorBoard (might throw and error).

yellowbee686: @Mazecreator  Thanks again, I'll try to figure out the TF summary mechanism and then modify my code
@AlexKuhnle @michaelschaarschmidt  I've notice that the default weight_init in Linear is Xavier of normal distribution
if self.weights_init is None:
    stddev = min(0.1, sqrt(2.0 / (x.shape[1].value + self.size)))
    self.weights_init = tf.random_normal_initializer(mean=0.0, stddev=stddev, dtype=tf.float32)
in the TF xavier_initializer, it doesn't have the min op, so could you explain the consideration about it, thanks!
https://www.tensorflow.org/api_docs/python/tf/contrib/layers/xavier_initializer
And also I think it's need MSRA initialization for PReLU according
https://arxiv.org/pdf/1502.01852.pdf

AlexKuhnle: @yellowbee686, the reason for this is that we at some point experimented with very small layer sizes in a toy example, which means that the `stddev` is rather big. We found that limiting it to `0.1` helped there. For "normal" layer sizes this restriction doesn't change the value at all, or at least not by much...

yellowbee686: @AlexKuhnle Thanks, that make sense, then I'll try to add the MSRA initialization and revert the episode reward code, which is apparently wrong with my misunderstand about TF summary.

yellowbee686:
close this PR for the default weight initialization is Xavier and episode_reward implement has errors

AlexKuhnle:
If you get the episode_reward to work, that would be very welcome. Otherwise we might look into it at some point.

yellowbee686:
@AlexKuhnle I'll try to make it work, and I create another PR for MSRA initialization

Mazecreator:
@AlexKuhnle and @yellowbee686 having played with the summary system, I don't think the current system will work for episode_reward as TF expects a value after each iteration.  I don't know enough about TF's template/monitored_session approach which seems to collect the summaries.  I bumped into this when I generated the "parameter scraping" to dump into TensorBoard.  In that case, I wanted to fire the summary exactly once, so I had to add this code https://github.com/reinforceio/tensorforce/blob/memory/tensorforce/models/model.py#L1126-L1131 to accomplish the task.

I think the calculation @yellowbee686  is doing for episode_reward is fine, but it isn't clear in the architecture how to record that value only on demand (when state is Terminal).  I wonder if the summary input variable can be change between "None & episode_reward" and have None reflect "DON'T RECORD"...

Just a thought...

AlexKuhnle:
@Mazecreator, you're right, this is a problem, and I realised the issue doesn't improve (rather the opposite) in the new memory branch, so we will have to think about a different system -- and maybe this means, after all, not using the summary writer functionalities of TensorFlow, since they seem to be very much designed for a straightforward supervised setup. We'll see what we can come up with...

PR #309, opened 2018-01-21
Beta old implementation

Mazecreator:
@michaelschaarschmidt & @AlexKuhnle ,

This is https://github.com/reinforceio/tensorforce/pull/298 rolled back but with spaces around "*" for PEP8.

Use this commit to avoid reported automated system error if it is believed to be real and we will need to figure out a better way to deal with the kwargs issues for the Nonlinear after commit.  I would recommend using 298 if the error is an artifact of the testing system.  It just isn't clear what environment is being used to test the code.


Mazecreator:
Obsolete, closing.

PR #326, opened 2018-02-05
Add MSRA weight initialization

yellowbee686:
I implement MSRA weight initialization for ReLU nonlinearity, according to https://arxiv.org/pdf/1502.01852.pdf ,
and the implementation refer to the mx.initializer.MSRAPrelu of mxnet.
Also I add a linear weight parameter from Dense pass to Linear.


michaelschaarschmidt:
Hey sorry for taking so long. Would you be able to rebase so this only includes the commits that are actually changed now? Because otherwise all the reverts will be included in the master history.

AlexKuhnle:
Did a squash merge, after all...

yellowbee686:
@michaelschaarschmidt @AlexKuhnle  Thanks, next time I'll take care of it.

PR #328, opened 2018-02-06
Unify Runner classes via BaseRunner parent class

sven1977:
Introduced BaseRunner class.
Restructured existing Runner classes (Runner and ThreadedRunner) to both inherit from BaseRunner.
Unified the interfaces used to control these classes and to handle per-episode and per-n-episode reporting.
Added multi-threaded testing to random and constant agent tests (configurable for all agents via run_type flag).
Removed some tf deprecation WARNINGS.

### DIFF
--- a/examples/threaded_ale.py
+++ b/examples/threaded_ale.py
@@ -175,7 +175,8 @@ def main():
 
     # Create runners
     threaded_runner = ThreadedRunner(
-        agents, environments,
+        agents,
+        environments,
         repeat_actions=1,
         save_path=args.save,
         save_episodes=args.save_episodes
@@ -183,9 +184,8 @@ def main():
 
     logger.info("Starting {agent} for Environment '{env}'".format(agent=agent, env=environments[0]))
     threaded_runner.run(summary_interval=100, episode_finished=episode_finished, summary_report=summary_report)
+    threaded_runner.close()
     logger.info("Learning finished. Total episodes: {ep}".format(ep=threaded_runner.global_episode))
-
-    [environments[t].close() for t in range(args.workers)]
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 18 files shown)


michaelschaarschmidt:
Looking good to me, could argue whether we should use multi threaded on an agent that is actually learning (vs random agent), but mergable. Don't know why rebasing does not work.
Any comments @AlexKuhnle ?

sven1977:
Ok, thanks guys!
I was trying to avoid adding more load to Travis (as Michael hinted in the gitter dev board). I will revert the 2 non-learning-agent tests as were (single mode) and add a new multi-threaded test case for PPO. Is there anything we can do about Travis accepting larger test loads? Like a paid account?

AlexKuhnle:
I think the tests recently take only something like 10-15min, and the limit is 45min or so (right, @michaelschaarschmidt ?), so that should be fine... (or I get something wrong with what kind of load was talked about^^).

michaelschaarschmidt:
I think there is a general travis issue where our tests terminate early, see recent builds, and it's unclear what exactly causes this (individual tests stalling/taking too many resources/build error from gym version)

sven1977:
Please hold on. Still testing PPO + multithreaded.
...

sven1977:
Ok, added another bug fix to multi-threaded functionality. All good now.
Added the multi-threaded run to the PPO agent. Each agent can now be configured via a bitfield to run in single, multi-threaded or both.

michaelschaarschmidt:
Does this mean we are ready to merge? We should merge this before merging memory in any case? @AlexKuhnle

sven1977:
My local tests were all ok. Not sure about travis. There seem to be new problems with openAI spaces and others. When I have some time, I'll take a look at this topic and try to stabilize the automatic tests.

michaelschaarschmidt:
I will merge this now but may make some changes on the tests (re: runmode), also in preparation for merging memory, because this definitely needs to be merged before, thanks!

sven1977:
Thanks Michael!

PR #340, opened 2018-02-18
Sequence and running_standardize preprocessors

sven1977:
Fixed sequence and running_standardize preprocessors in memory branch.
Bug fix: model.reset() has to return dict instead of list for self.internals_init.
Fixed counter some stats for Runner/ThreadedRunner.
Fixed some comments (e.g. network vs network_spec) due to new variable naming scheme.


sven1977:
Thanks Alex!

AlexKuhnle:
Thanks to you, @sven1977!  ;-)

PR #358, opened 2018-03-04
internal_lstm: allow passing arbitrary args to LSTMCell()

lefnire:
For consideration. I'm experimenting w/ LSTMCell's `cell_clip=3` param as a way to mitigate _very_ long time-series sequences (vanishing/exploding gradient) from something [I'd read](https://arxiv.org/pdf/1504.01482.pdf). Grain of salt, I'm no expert, so LMK if I'm barking up the wrong tree. Anyway, I'm using this commit https://github.com/lefnire/tensorforce/commit/c33a2648702f6efb29549c98a0adacb47edc291d so I can hyper-search over `cell_clip=[bounded]` & `use_peepholes=[bool]`. I'm thinking maybe it'd just be better to have `lstmcell_args=dict()` which can take any arg LSTMCell would take.

### DIFF
--- a/tensorforce/core/networks/layer.py
+++ b/tensorforce/core/networks/layer.py
@@ -1080,7 +1080,7 @@ class InternalLstm(Layer):
     Long short-term memory layer for internal state management.
     """
 
-    def __init__(self, size, dropout=None, scope='internal_lstm', summary_labels=()):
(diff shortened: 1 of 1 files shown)


michaelschaarschmidt:
That seems fine to me, thanks

PR #366, opened 2018-03-11
Mentioning minimum TF version for  Tensorforce

HassamSheikh:
The sampling_body method in replay.py uses tf.while_loop and is given a parameter named as maximum_iterations. This parameter is introduced in Tensorflow version 1.5.  If Tensorflow 1.4 is used for DQN and similar agents, we get unknown args 'maximum_iteration' error


michaelschaarschmidt:
Thanks!

PR #369, opened 2018-03-14
Added possibility to specify tensorflow session config for all session types not only distributed

philipshurpik:
Hi!

Thanks for your project.
I started using your project but found that I'm unable to specify tensorflow session config when doing not distributed training.

I found that it was done before in https://github.com/reinforceio/tensorforce/pull/237
But unfortunately now this code does not exist, and it's impossible to set session_config from outside of the library.

Also, I moved session_config from the distributed spec because it's not only distributed functionality (and to not duplicate it in arguments)

### DIFF
--- a/tensorforce/models/memory_model.py
+++ b/tensorforce/models/memory_model.py
@@ -47,7 +47,8 @@ class MemoryModel(Model):
         update_mode,
         memory,
         optimizer,
-        discount
+        discount,
+        session_config=None
     ):
         """
         Memory model.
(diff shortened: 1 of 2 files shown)


michaelschaarschmidt:
Hm,
should this not then also be part of the agent API? I think long-term I would make this part of an execution config dict handling these aspects, but for now it could be either part of distributed spec or a separate agent constructor parameter. I agree it's not distributed specific.

michaelschaarschmidt:
Have begun to do execution refactoring to have a high level execution config, session config is now part of that so that addresses your issue

philipshurpik:
thanks!

michaelschaarschmidt:
Closing then

PR #370, opened 2018-03-14
Independent save and restore of model components

janislavjankov:
This change introduces the mixin `SavableComponent` that model
components that are to be saved or restored independently from the
model should extend from.

The Model calls `register_saver_ops` of each savable component when
computation graph is constructed and save operations can be added to
it.

Note that with this change the save and restore of the savable
components need to be called explicitly. Model's save, restore and
checkpoint functionality does not delegate to the savable components'
save and restore methods.

A simple recipe to train/preload embeddings can be found in
TestModelSaveRestore.test_pretrain_network

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -134,3 +134,82 @@ actions = agent.act(states=state)
 In the case of multiple actions, the values will be arrays. It is a good idea
 to use expressive action names within your application to keep track of 
 each action and its purpose.
+
+### 7. How do I preload values in my model?
+
+There are scenarios where you might want to initialize the model or parts of the model with values
+learnt from training on another task or completely outside of the Tensorforce framework. Here is a
+simple recipe of how to do this:
+
+1) You need to define the component of the model that you want to save or restore as
+`SavableComponent`. The example below adds the `SavableComponent` mixin to `LayeredNetwork` so that
+the network component can be saved and restored independently.
+
+   ```python
+   from tensorforce.util import SavableComponent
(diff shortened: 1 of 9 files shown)


michaelschaarschmidt:
So one thing we could think about is whether we can avoid making calls like 
```agent.model.network.restore(sess=agent.model.session, save_path=save_path) ```

and have an agent-level API function to perform these things, ideally throwing useful errors if one attempts to restore components which do not have this functionality. What do you think?

**Comment by janislavjankov** [comment]
Makes sense, it will hide some of the details from the caller. For instance we can add this to the model:
```
restore_network(save_path):
    """
    Restores the network component using this model's session
    Throws:
        TensorforceError if the network is not a SavableComponent
    """
    pass

save_network(save_path):
    """
    Throws:
        TensorforceError if the network is not a SavableComponent
    """
    pass
```

My only concern here is that we'll need this pair of methods for every component of the model. Alternative is to use some form of named components, i.e. `restore_component(component_name, save_path)`, where the component name is "network" or "target_network" or "distribution", etc depending on the specific model. Let me know and I can go ahead with the changes.

michaelschaarschmidt:
Yes I was thinking more along the lines of passing the component name. If you look at the act() api, there is a fetch_tensors option where you can pass names of tensors to fetch, and it seems like this could have a similar structure. I think this functionality should also be explained in a few sentences somewhere, maybe the FAQ? Or in the agent api method.

michaelschaarschmidt:
Looks good, thank you!

PR #377, opened 2018-03-21
Further cleanup of model.py:setup().

sven1977:
Major refactor in model.py.initialize: Split, renamed and further packaging of sub-tasks by the model's setup function. This will make it easier to write sub-classes for `Model`. Added some missing doc-strings. 
Moved global counters into a cleaner tf.get_variable-logic (instead tf.Variable).

Slight Memory-classes refactor. Moved tf function creation into its own setup function called by init.

Optimizers: Some commenting, removed get_wrapper (replaced by functools.partial).

### DIFF
--- a/tensorforce/contrib/sanity_check_specs.py
+++ b/tensorforce/contrib/sanity_check_specs.py
@@ -129,7 +129,7 @@ def sanity_check_execution_spec(execution_spec):
 
     # TODO: Figure out what exactly we need for options and what types we should support.
     if type_ == "distributed":
-        def_ = dict(job="ps", task=0, cluster_spec={
+        def_ = dict(job="ps", task_index=0, cluster_spec={
             "ps": ["localhost:22222"],
             "worker": ["localhost:22223"]
         })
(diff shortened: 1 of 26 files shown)


michaelschaarschmidt:
What is the reasoning for 

'''-    def tf_step(self, time, variables, global_variables, **kwargs):'''

removing global_variables as an explicit argument?

sven1977:
Pycharm complained that the signature did not match the parent's signature (see: Liskov substitution principle):

"Signature of method 'SomeClass.func()' does not match signature of base method in class 'SomeClass' less... (Ctrl+F1) 
This inspection detects inconsistencies in overriding method signatures."

michaelschaarschmidt:
No that makes sense, and I agree listening to Pycharm warnings is usually a very good idea

sven1977:
Ok, cool.
However, again, in this case we are violating Liskov substitution principle either way, due the global_variables being enforced. :|

sven1977:
Thanks Michael!

PR #380, opened 2018-03-26
OpenAiGym integration bugfix

mannsi:
Fix integration bug caused by an interface change in OpenAiGym multidiscrete space. Was trying to access properties of the space that do not exist. Bug was discovered against OpenAiGym 0.10.4.


michaelschaarschmidt:
thanks!

PR #385, opened 2018-04-02
Finish API changes to remove _spec

nrhodes:
Commit reinforceio/tensorforce@ec7c970 modified some
agent/model arguments.

This commit fixes some uses of the API that were still using the
old argument names

Changed:
  states_spec to states
  actions_spec to actions
  network_spec to network


michaelschaarschmidt:
Thanks!

Issue #388, opened 2018-04-04
Large memory.capacity value causes tensorflow type issues


trickmeyer:
Running the PPOAgent with all values set the same as in the ppo json file example, but memory.capacity set high causes a type check failure.

Here's an example snippet, where only memory.capacity is changed from the json example:

```
        memory=dict(
                type='latest',
                include_next_states=False,
                capacity=(1000000)
            )

```

And here's the error (truncated to the last Tensorforce call):

```
  File "/home/tom/temp/tensorforce/tensorforce/core/memories/memory.py", line 64, in custom_getter
    variable = getter(name=name, registered=True, **kwargs)
  File "/home/tom/temp/tensorforce/tensorforce/models/model.py", line 501, in custom_getter
    variable = getter(name=name, **kwargs)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/ops/variable_scope.py", line 408, in _true_getter
    use_resource=use_resource, constraint=constraint)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/ops/variable_scope.py", line 800, in _get_single_variable
    use_resource=use_resource)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/ops/variable_scope.py", line 2157, in variable
    use_resource=use_resource)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/ops/variable_scope.py", line 2135, in <lambda>
    return lambda **kwargs: captured_getter(captured_previous, **kwargs)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/ops/template.py", line 384, in _checkpointable_custom_creator
    **kwargs)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/training/checkpointable.py", line 415, in _add_variable_with_custom_getter
    **kwargs_for_getter)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/ops/template.py", line 371, in _call_next_creator_renaming_initializer
    **inner_kwargs)
  File "/home/tom/miniconda2/envs/tf2/lib/python2.7/site-packages/tensorflow/python/ops/variable_scope.py", line 2135, in <lambda>
[...]

PR #393, opened 2018-04-06
Initialize terminal memory to match tensorflow's expected type

trickmeyer:
Fix for #388.  When queue.capacity is an np.int64, the comparison leads to an np.bool_ type, rather than a python bool, which is what's necessary for the initializer, which is a tf.bool.

### DIFF
--- a/tensorforce/core/memories/queue.py
+++ b/tensorforce/core/memories/queue.py
@@ -102,7 +102,7 @@ class Queue(Memory):
             shape=(self.capacity,),
             dtype=util.tf_dtype('bool'),
             initializer=tf.constant_initializer(
-                value=tuple(n == self.capacity - 1 for n in range(self.capacity)),
+                value=[n == self.capacity - 1 for n in range(self.capacity)],
                 dtype=util.tf_dtype('bool')
             ),
             trainable=False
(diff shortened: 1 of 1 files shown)


trickmeyer:
I actually figured out this is something else unrelated, but potentially important.  I have an internal library I use to randomly explore hyperparameters, and it uses np.random, which returns np types.  When those are fed as inputs into many different places into PPOAgent's `__init__` it causes type inconsistencies.  

I can recreate by passing in 10,000 as memory.capacity as `np.int64` vs. passing it in as a python int.  Former fails (w/o this PR), latter succeeds.

It's likely not worth doing this one fix, it's a bigger issue with `np.int` inputs.

michaelschaarschmidt:
I think this is something about how we do type mapping in util - https://github.com/reinforceio/tensorforce/blob/master/tensorforce/util.py#L60 
-> missing 64 bit types? Thanks for the fix!

trickmeyer:
I was looking at your utils, but didn't have time to chase the full pipeline down.  You can recreate by passing in an np.int64 into any argument normally passed in as a simple python int.  

I was thinking the best way to handle this would be to check/convert in the base (or lowest) `Agent` and `Model` classes that are the ultimate receivers of the config values, but that doesn't handle arguments specific to a certain `Agent` or `Model`.  For now, just keep in mind anytime someone gets a tensorflow expected type error, this is likely the cause.

PR #406, opened 2018-05-03
added gym.spaces.Dict as an available action/state space for OpenAIGy…

### LINKED ISSUE #405: OpenAIGym Environment does not support Dict/nested spaces
I built a gym environment for a problem I'm working on. My action space looks like:
```python
import gym
from gym import spaces

class MultipleMoverGridWorld(gym.Env):
    ...
    def __init__(self):
        self.action_space = spaces.Dict({
        "mover{}".format(i): spaces.Discrete(5) \
                for i in range(self.n_movers)
        })
```
with each mover in my grid world having 5 possible actions (up/down/left/right/null). My environment allows multiple movers to change position each time step. I get the following error when trying to get the tensorforce actions:
```
---------------------------------------------------------------------------
TensorForceError                          Traceback (most recent call last)
<ipython-input-515-20ed75416bda> in <module>()
----> 1 OpenAIGym.action_from_space(space=env.action_space)

~\AppData\Local\ESRI\conda\envs\cpd\lib\site-packages\tensorforce\contrib\openai_gym.py in action_from_space(space)
    150             return actions
    151         else:
--> 152             raise TensorForceError('Unknown Gym space.')

TensorForceError: Unknown Gym space.
```

Near as I can tell, I should be able to implement the functionality in tensorforce as:
```
actions = {
    "mover{}".format(i): dict(type='int', num_actions=5) \
            for i in range(self.n_movers)
}
```

I can definitely work on a PR for this if there's interest, but definitely want to check if there's a workaround first. I have my environment implemented as an OpenAI gym and would prefer not to reimplement as a tensorforce environment if I have the option. :)

EDIT: I should also add that this is only a minimal example of this: my action space will be a little more complex in the future.

dwilson1988:
added gym.spaces.Dict as an available action/state space for OpenAIGym environments. Partially fixes #405. Still need to investigate the effects of nested spaces within the Environment class

### DIFF
--- a/tensorforce/contrib/openai_gym.py
+++ b/tensorforce/contrib/openai_gym.py
@@ -101,6 +101,13 @@ class OpenAIGym(Environment):
                         states['state{}'.format(n)] = state
                         n += 1
             return states
+
+        elif isinstance(space, gym.spaces.Dict):
+            states = dict()
+            for space_name,space in space.spaces.items():
+                state = OpenAIGym.state_from_space(space=space)
+                states[space_name] = state
+            return states
         else:
             raise TensorForceError('Unknown Gym space.')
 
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Once you investigated recursive space definitions, a PR covering these would of course be very welcome... :-)

PR #409, opened 2018-05-06
Add a saving and visualization functionality to OpenAI Gym's example

coord-e:
Added options to save trained models, and to enable the visualization of OpenAI Gym.


michaelschaarschmidt:
Thanks!

PR #412, opened 2018-05-13
Also pass shape to single preprocessers not just stacked ones

erniejunior:
Otherwise the constructor of Preprocessor complains that the kwarg 'shape' is not set.


michaelschaarschmidt:
Thanks!

PR #419, opened 2018-05-21
Implement testing flag to runner to evaluate agent with runner.

benelot:
A simple implementation to allow the runner to test an agent on an environment without learning. Loading weights from file and running the agent on an environment is then simply like the enjoy_scripts in openai/baselines.

P.S. Thanks for this library, is is absolutely amazingly clean and well designed. I would love to contribute to this in the future.

I implemented the test flag in openai_gym.py for now, but it can easily be added to all other examples as soon as you guys are happy with my idea.


michaelschaarschmidt:
Thanks!

PR #422, opened 2018-05-23
Fix for restore from checkpoint

janislavjankov:
Resets the buffer index upon restore, ensuring the model is consistent
with the agent's behavior - starting a new episode.

This fixes the following error which occurs on attempt to restore:
    InvalidArgumentError (see above for traceback): Must have updates.shape = indices.shape + params.shape[1:], got updates.shape [30,1], indices.shape [21], params.shape [100,1]
  	[[Node: ppo/observe-timestep/store/ScatterUpdate = ScatterUpdate[T=DT_FLOAT, Tindices=DT_INT32, _class=["loc:@ppo/initialize/latest/initialize/state-state"], use_locking=true, _device="/job:localhost/replica:0/task:0/device:CPU:0"](ppo/initialize/latest/initialize/state-state, ppo/observe-timestep/store/mod, ppo/strided_slice, ^ppo/observe-timestep/store/AssignSub)]]

### DIFF
--- a/tensorforce/models/model.py
+++ b/tensorforce/models/model.py
@@ -216,6 +216,8 @@ class Model(object):
         self.timestep_output = None
 
         self.summary_configuration_op = None
+        # Add an explicit reset op with no dependencies
+        self.buffer_index_reset_op = None
 
         # Setup Model (create and build graph (local and global if distributed), server, session, etc..).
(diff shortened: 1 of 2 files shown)


michaelschaarschmidt:
Thanks!

PR #424, opened 2018-05-24
Implement sleep parameter to runner to allow slowing down execution

benelot:
This is especially useful when analyzing the visualization of performance. Running it at normal speed is usually too fast to see what is going on. Like this, it is possible to set --sleep x (seconds, e.g. 0.05 ) in order to make the runner sleep before continuing.

If you do not like the scaling of the sleep representation of time (x vs. 1/x), we can change that to something more useful. I just thought it might make sense to allow the user to choose arbitrary numbers (never restrict too much if you do not know what people are doing with it). We can discuss this here if necessary.


michaelschaarschmidt:
A bit specific but why not - thanks

benelot:
Do you mean specific because it is just helpful in one specific case? I
understand, but I think it makes the runner an object to simplify quick
experimentation. Thanks for merging!

Issue #433, opened 2018-06-14
Memory capacity not aligned with episode length could result in InvalidArgumentError


janislavjankov:
When the memory capacity is not perfectly aligned with the episode length, retrieving the episodes from memory can result in empty responses when data was previously stored. This causes InvalidArgumentError downstream like the following:
```
InvalidArgumentError (see above for traceback): Need minval < maxval, got 0 >= 0
	 [[Node: ppo/observe-timestep/cond/optimization/multi-step/step/subsampling-step/step/random_uniform = RandomUniformInt[T=DT_INT32, Tout=DT_INT32, seed=0, seed2=0, _device="/job:localhost/replica:0/task:0/device:CPU:0"](ppo/observe-timestep/cond/optimization/multi-step/step/subsampling-step/step/random_uniform/shape, ppo/observe-timestep/cond/optimization/multi-step/step/while/Const, ppo/observe-timestep/cond/optimization/multi-step/step/subsampling-step/step/strided_slice)]]
```

or

```
InvalidArgumentError (see above for traceback): Inputs to operation trpo/observe-timestep/cond/optimization/reward-estimation/Select of type Select must have the same size and shape.  Input 0: [0] != input 1: [1]
	 [[Node: trpo/observe-timestep/cond/optimization/reward-estimation/Select = Select[T=DT_FLOAT, _device="/job:localhost/replica:0/task:0/device:CPU:0"](trpo/observe-timestep/cond/optimization/reward-estimation/Select/Switch:1, trpo/observe-timestep/cond/optimization/reward-estimation/zeros_like, trpo/observe-timestep/cond/optimization/reward-estimation/concat)]]
```

I've added a test that demonstrates the problem here: https://github.com/janislavjankov/tensorforce/blob/fix/memories/tensorforce/tests/test_memory.py#L112
PR is on the way

PR #435, opened 2018-06-17
Fix the queue-based memories

janislavjankov:
This addresses issue #433. 

The episode_indices array was not properly updated leading to under-utilization of the buffer and inaccurate episode and timesteps retrieval when the capacity was not perfectly aligned with the episode length.


michaelschaarschmidt:
Thanks!

Issue #438, opened 2018-07-01
Environment Documentation Not Consistent for Execute(actions)


daggertye:
Hey,

I was going over the source code and testing out environments. In particular, I tried running a list of actions 

```
env = OpenAIGym("CartPole-v0")
env.execute({'action0': 0, 'action1': 1})
```

fails to run. Going over the source code and stack trace for OpenAIGym reveals that it attempts to parse it in the code block

```
if isinstance(actions, dict):
            actions = [actions['action{}'.format(n)] for n in range(len(actions))]
```

but openai gym doesn't support stepping with a list of actions. Going over the documentation and various implementation for environments reveals further inconsistencies. In particular, the signature and documentation for Environment.execute 

```
def execute(self, actions):
        """
        Executes action, observes next state(s) and reward.

        Args:
            actions: Actions to execute.

        Returns:
            (Dict of) next state(s), boolean indicating terminal, and reward signal.
        """
```

hints at being able to execute on multiple actions, but the implementations for Game2048 and Ale seem to hint at actions only being singular. But GymOpenAI and OpenAIUniverse hint at multiple options. Therefore, I must ask for some clarification on what exactly is the behavior. 

Thanks

daggertye:
I fix this in #448 and only single actions can be given.

PR #439, opened 2018-07-01
pygame_learning_environment env implemented

daggertye:
Great project! After going over the source code, I feel that the project is really nicely built. I've decided to try and go ahead an implement the pygame_learning_environment to get familiarized with the api. 

My main takeaways are that the api is somewhat confusing (issue #438) and also that some libraries (doom) are deprecated.


daggertye:
Nevermind, something's wrong.

AlexKuhnle:
Thanks for integrating!

Issue #442, opened 2018-07-03
OpenAI/Universe environment deprecated


daggertye:
Hey,

The OpenAI universe [repository](https://github.com/openai/universe) is archived and is deprecated for python 3. Instead, they suggest [Retro](https://github.com/openai/retro). Can we provide support for this library as well?

AlexKuhnle:
In principle, sure. Are you free to look into and contribute the interface? Otherwise might take some time, just getting started with support again.

PR #457, opened 2018-07-25
OpenAIRetro implementation

daggertye:
in #442 I raised the issue of OpenAi/universe being deprecated in favor of retro. Here is integration for the new environments.


AlexKuhnle:
That was quick, thanks! :-)
One comment: I think you also need to add the un-/flatten operations that I added yesterday for Gym, in 75c4c93 and e8ffab7. If possible, it would also be great to have a clean PR without the big merge, but that's secondary.

daggertye:
Ok, I'll close this for now.

PR #462, opened 2018-07-29
Contributed Environments normalized + base environment class docs expanded

daggertye:
This commit expands on the environment class and clarifies what is returned.


AlexKuhnle:
Thanks for the cleanup and docu!

PR #479, opened 2018-08-21
minor change to the sample command to run openai_universe.py

fqhuy:
minor change to the sample command to run examples/openai_universe.py


AlexKuhnle:
Thanks for updating.

PR #497, opened 2018-09-11
Implemented Dense(trainable=False)

batu:


batu:
From the other requests it seem that the test isn't very reliable. Reopening.

AlexKuhnle:
There's one thing missing: passing on the `trainable` parameter in the constructor of `Dense` as an argument to `Linear`. Minor thing: could you also move the `trainable` argument(s) before the `named_tensors` argument in the argument list? (The argument list starts with more layer-specific arguments and ends with some general internal ones.)

AlexKuhnle:
Thanks!

PR #503, opened 2018-09-15
Minor bug fix in Linear Layer and Epsilon anneal exploration

### LINKED ISSUE #504: AttributeError: 'Linear' object has no attribute 'trainable'
I get this exception when I try to run DQN agent. I made an inspection and found out that "trainable" property is not set in the Linear Layer. I have updated the code and created a pull [request.](https://github.com/reinforceio/tensorforce/pull/503)

Jonathan-Livingston-Seagull:
This fixes #504 and #506


AlexKuhnle:
Thanks for the fix! Actually, is this `tf.fill(...)` fix necessary for `constant` exploration as well?

PR #505, opened 2018-09-16
An implementation of KFac Optimization compatible with tensorforce

### LINKED ISSUE #183: Implement KFAC-optimizer
Necessary for ACKTR but would also be available for other models via the new optimizers.

daggertye:
Closes #183

### DIFF
--- a/tensorforce/core/optimizers/kfac.py
+++ b/tensorforce/core/optimizers/kfac.py
@@ -1,0 +1,966 @@
+# Copyright 2018 reinforce.io. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


AlexKuhnle:
hey @daggertye, that sounds really cool! Have you experimented with it already? I scrolled over it and found one likely tpo in line 926 regarding `var` and `vars`. Apart from that, and assuming you did some testing, I'm happy to merge. Great work!

AlexKuhnle:
Could you run some basic checks whether it learns something?

daggertye:
I ran it on a couple of networks and it was optimizing the loss, as desired.

AlexKuhnle:
Cool, I assume that we can merge then. Thanks for the great work!

PR #508, opened 2018-09-18
Implementation of parallel evaluation of actions and observations

Deathn0t:
Corresponding to the execution config (with num_parallel) of an agent :
Tested with PPOagent, with num_parallel=1 and num_parallel=2, and with update_mode(batch_size=1, frequency=1) and update_mode(batch_size=10, frequency=10)
```
execution=dict(
            type='single',
            num_parallel=2,
            session_config=None,
            distributed_spec=None)
```

### DIFF
--- a/examples/quickstart.py
+++ b/examples/quickstart.py
@@ -52,9 +52,9 @@ agent = PPOAgent(
     update_mode=dict(
         unit='episodes',
         # 10 episodes per update
-        batch_size=20,
+        batch_size=10,
         # Every 10 episodes
-        frequency=20
+        frequency=10
     ),
     memory=dict(
         type='latest',
--- a/tensorforce/agents/agent.py
+++ b/tensorforce/agents/agent.py
@@ -63,10 +63,6 @@ class Agent(object):
         # Batched observe for better performance with Python.
         self.batched_observe = batched_observe
         self.batching_capacity = batching_capacity
-        if self.batched_observe:
-            assert self.batching_capacity is not None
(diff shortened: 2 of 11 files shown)


AlexKuhnle:
Hey @Deathn0t, thanks very much for contributing this feature. Looks all good!

sam-ngu:
Hi, thanks a lot for this feature!
I noticed there is an issue in 'parallel_runner.py' line 137
``` python 

                self.episode_actions.append(self.environment.conv_action)
```
Correct me if I'm wrong, but as far as i know, the environment class does not have the conv_action property. 

What does conv_action stand for? I'm guessing you are trying to keep track of all the actions taken by the agent in the current episode?

AlexKuhnle:
Hey @sam-ngu, letting @Deathn0t answer the specific question, but just to let you know that there is a new version coming which you can find under the `major-revision` branch. There are a few API changes, like `num_parallel` is called `parallel_interactions` and is a high-level argument, but it also supports an updated `parallel_runner` feature ([see here](https://github.com/tensorforce/tensorforce/blob/major-revision/tensorforce/execution/parallel_runner.py)). So you may consider switching, and happy to answer porting questions e.g. on Gitter.

sam-ngu:
@AlexKuhnle Thanks a lot for the info! Will check it out.

Deathn0t:
Hi @sam-ngu , you're right this is a mistake ! I was using a specific environment which was having this `conv_action` attribute that's why!

PR #509, opened 2018-09-19
Add support to OpenSim-RL

andrewekhalel:
Reinforcement learning with musculoskeletal models in OpenSim
http://osim-rl.stanford.edu/


AlexKuhnle:
@andrewekhalel, looks great, thanks for contributing!

Issue #513, opened 2018-09-23
Possible bug running DDPG


sohakes:
Hi!

I was trying to make DDPG work, but there is a bug or a error in my side. I have an environment with 21 continuous observations and 3 continuous actions, I also made the batch of size 2 (just for testing). Every time I tried to run, I encountered this error after some timesteps:

`InvalidArgumentError (see above for traceback): ConcatOp : Dimensions of inputs should match: shape[0] = [2,32] vs. shape[1] = [6,1]
     [[Node: ddpg/observe-timestep/cond/optimization/critic/apply/concat = ConcatV2[N=2, T=DT_FLOAT, Tidx=DT_INT32, _device="/job:localhost/replica:0/task:0/device:CPU:0"](ddpg/observe-timestep/cond/optimization/critic/apply/relu0/apply/Relu, ddpg/observe-timestep/cond/optimization/critic/apply/Reshape, ddpg/observe-timestep/cond/optimization/critic/apply/concat/axis)]]`

In the traceback, the error was apparently in this code https://github.com/reinforceio/tensorforce/blob/master/tensorforce/models/dpg_target_model.py, in the line:

`out = self.t0l.apply(x=x_states, update=update)`

I tried to understand what's happening before posting the issue, but I may have found a solution. I thought that, in `tf_apply`, the idea is to connect the observations to the first layer, and then concatenate the first layer with the actions. But since there was a reshape of the actions, instead of size [actions, batch] we had size [actions*batch, 1], and so it wasn't possible to concatenate.

I end up removing the reshape and making the concatenation on the axis=1 so it concatenate each observation of the batch. It's running without error now, but honestly I'm too lost to know if I did it right, just disguised the error, or if it's something else in my configuration that I did wrong. The reward is not improving, but it also isn't with PPO, so it may be my environment.

Thanks

AlexKuhnle:
Unfortunately I'm not too sure about the current implementation of DDPG, but it definitely needs cleaning up (and maybe fixing). Can you create a PR with the change you made, which at least lets it run without exception? I can't say too much about whether something is wrong with DDPG, only that PPO generally works quite well, so it may indicate that something is wrong with the environment.

sohakes:
Thanks for answering. Since opening this issue, I've tried running again with the default exploration (I was trying with Ornstein Uhlenbeck) and now the reward in both DDPG and PPO is increasing slowly, so the problem with the rewards may be due to that.

[...]

PR #514, opened 2018-09-23
Fixes a bug with DDPG

sohakes:
It's the change I made regarding issue https://github.com/reinforceio/tensorforce/issues/513.


AlexKuhnle:
Thanks!

Issue #515, opened 2018-09-23
Custom critic network support for DDPG


Jonathan-Livingston-Seagull:
DDPG does not support custom critic network in the current implementation.

AlexKuhnle:
Resolved.

PR #516, opened 2018-09-23
added support for ddpg custom critic network

Jonathan-Livingston-Seagull:
Fix for #515 .


AlexKuhnle:
Thanks a lot, @Jonathan-Livingston-Seagull !

PR #520, opened 2018-10-07
roboschool support

s-udhaya:
Few environments of gym needs Mujoco installation. Mujoco provides free license only for students. This pull request adds support for roboschool which replicates gym Mujoco dependent environments using bullet physics engine and does not need license.


AlexKuhnle:
I think this feature duplicates what the `-i` command line argument ([here](https://github.com/reinforceio/tensorforce/blob/master/examples/openai_gym.py#L47)) tries to solve, doesn't it?

PR #524, opened 2018-10-12
Support for Unity ML Agents

vermashresth:
Hi!
I have implemented bindings for Unity environments. This requires Unity mlagents v0.5
Some things that I couldn't figure out
1. Unity provides running multiple instances of the same environment so as to collect data parallelly. But this means collecting observations from all of them and passing action to all of those instances. That's like a batch of environments. Any idea how to implement that? Currently, I'm throwing `unsupported` error in these cases.
2. Unity allows having multiple `brains` (multi-agent environment) each requiring its own training network. Can multiple agents be trained simultaneously using Tensorforce?
3. In some environments, there can be visual observations along with vector observations, so how to pass on both of them together?
4. For the example script, I have included prebuilt binary of a Unity environment, but that is 75 MB already so maybe it is a bad idea?


AlexKuhnle:
Thanks for the PR, here some comments:

1. There are various features in TensorForce that may be applicable: using distributed TensorFlow, using a [threaded runner](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/execution/threaded_runner.py), or using parallel episodes ([index argument](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/agents/agent.py#L104)). I can't say which would work best for what Unity provides here...
2. It's not really possible, I think, to train multiple agents within the same instance of TensorForce/TensorFlow (within the same process etc). In principle it should be, but that would be quite some work.
3. The states specification makes it possible to specify multiple states / state components, so you could specify the visual input and the vector as another input with a different shape. That requires a [more complex network specification](https://github.com/reinforceio/tensorforce/blob/master/tensorforce/tests/base_agent_test.py#L134).
4. Yes, I would remove the environments and just add comments and command line arguments in the unity script to explain how to use an environment.

AlexKuhnle:
@vermashresth  is this still something you want to look at?

vermashresth:
@AlexKuhnle  HI! I'm a bit caught up with my university exams these days but I would like to try it after a week or two.

AlexKuhnle:
Okay, no problem, just wanted to check. :-)

AlexKuhnle:
Outdated.

PR #525, opened 2018-10-18
fixed index reset op

Jonathan-Livingston-Seagull:
list_buffer_index_reset_op was not initialized as tensor, hence created problem while indexing.


AlexKuhnle:
Thanks!

PR #580, opened 2019-03-29
fix environments tips

kaixinbaba:
environments to environment


AlexKuhnle:
Hehe, makes the message more helpful, for sure. Thanks!

PR #616, opened 2019-10-04
fix Agent seeding

christianhidber:
fixes typo in Agent.__init__ resulting in an exception when seed is not None.
added unit test to test_agent, which simply instantiates an agent with a non-trivial seed

question: is this the right place for a unit test like this ? is there a better way to test the instantiation with a seed ?


AlexKuhnle:
Many thanks!

(FYI: May move the unittest to a separate file at some point, because there are probably still a few other "optional" arguments which need coverage.)

PR #619, opened 2019-10-14
Fixing registration bug where level is concatenated with a string before it's converted TO a string.

### LINKED ISSUE #618: 0.5.1 OpenAIGym unable to create level from custom environment
OpenAIGym#create_level() has an error when a custom gym has not been registered in the gym environment. The code is attempting to concatenate `level + '-v0'` in order to create a name to use to iterate over the env.registry but level is an object and not a string.

I've created a bugfix on a local fork which I will append to the ticket.

codeninja:
fixes #618

### DIFF
--- a/tensorforce/environments/openai_gym.py
+++ b/tensorforce/environments/openai_gym.py
@@ -66,8 +66,8 @@ class OpenAIGym(Environment):
 
         # Find level
         if level not in gym.envs.registry.env_specs:
-            if level + '-v0' in gym.envs.registry.env_specs:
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Hi, thanks for the PR. I'll look into it later today.

AlexKuhnle:
I think this problem has been addressed with a recent change on master. I will update the version soon, would be great if you could test it again then.

codeninja:
This problem has been resolved for me. Closing this PR.

PR #621, opened 2019-10-18
Fix global variable naming in model.py

rshtirmer:
Fix global variable naming in model.py. I got an error about this issue when making a "flatten" layer, and this fixes the issue.

### DIFF
--- a/tensorforce/core/models/model.py
+++ b/tensorforce/core/models/model.py
@@ -686,7 +686,7 @@ class Model(Module):
                 x=self.global_episodes, operation_name='episode-output'
             )
             update = util.identity_operation(
-                x=self.global_update, operation_name='update-output'
+                x=self.global_updates, operation_name='update-output'
             )
 
         return timestep, episode, update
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Hi, could you give more information about the error you got? As far as I can tell, this renaming shouldn't work. (See the name of the global tensors e.g. [here](https://github.com/tensorforce/tensorforce/blob/master/tensorforce/core/module.py#L345))

rshtirmer:
Interesting, the error was essentially that self.global_episode does not exist, which went away after renaming.

AlexKuhnle:
Hmm, so it doesn't need fixing after all?

rshtirmer:
@AlexKuhnle Well, no... the changes I made were renaming... which is what fixed it.. If you'd like I can get you a copy of the error log.

AlexKuhnle:
Hmm, that is weird, since the names of the variables seem to be "singular", so `global_episode` etc, as can be seen in the link above, but your changes suggest they should be in plural, `global_episodes`. A copy of the exception would be very helpful. Also, just to check: Are you on the latest master version?

AlexKuhnle:
I will close this PR due to inactivity.

PR #626, opened 2019-10-29
Socket remote env

jerabaul29:
@AlexKuhnle this is for pushing as a contribution the socket wrapper classes. Of course this is a bit user-dependent, but hope this can help. This is a slightly more recent setup than from the paper.

### DIFF
--- a/tensorforce/contrib/socket_remote_env/.gitignore
+++ b/tensorforce/contrib/socket_remote_env/.gitignore
@@ -1,0 +1,3 @@
+env_*
+__pycache__
+saver_data
--- a/tensorforce/contrib/socket_remote_env/README.md
+++ b/tensorforce/contrib/socket_remote_env/README.md
@@ -15,3 +15,14 @@ See the article, and the code online, for further help:
 ```
 https://github.com/jerabaul29/Cylinder2DFlowControlDRLParallel
 ```
+
+NOTE: this is tested only on Ubuntu 18.04
+
+To test, either:
+
+- use our script: ```bash script_launch_parallel.sh  training_example 3000 4```
+
+- launch by hand:
+
+  - ```python3 launch_servers.py -p 3000 -n 4```
+  - ```python3 launch_parallel_training.py -p 3000 -n 4```
--- a/tensorforce/contrib/socket_remote_env/README.md
+++ b/tensorforce/contrib/socket_remote_env/README.md
@@ -1,0 +1,17 @@
+This code allows to wrap the environment through a socket for training through the network. For example, this allows
+to run environments on a set of machine, and the Agent on a separate machine.
+
+This code was developped originally in:
+
+```
+Rabault, J., Kuhnle, A (2019).
+Accelerating Deep Reinforcement Leaning strategies of Flow Control through a
+multi-environment approach.
+Physics of Fluids.
+```
+
+See the article, and the code online, for further help:
+
+```
+https://github.com/jerabaul29/Cylinder2DFlowControlDRLParallel
+```
--- a/tensorforce/contrib/socket_remote_env/README.md
+++ b/tensorforce/contrib/socket_remote_env/README.md
@@ -18,7 +18,7 @@ https://github.com/jerabaul29/Cylinder2DFlowControlDRLParallel
 
 NOTE: this is tested only on Ubuntu 18.04
 
-To test, either:
+To test / run the example, either:
 
 - use our script: ```bash script_launch_parallel.sh  training_example 3000 4```
 
--- a/tensorforce/contrib/socket_remote_env/RemoteEnvironmentClient.py
+++ b/tensorforce/contrib/socket_remote_env/RemoteEnvironmentClient.py
@@ -1,0 +1,129 @@
+from tensorforce.environments import Environment
+import socket
+from echo_server import EchoServer
+
+
+class RemoteEnvironmentClient(Environment):
+    """Used to communicate with a RemoteEnvironmentServer. The idea is that the pair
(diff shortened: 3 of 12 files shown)


jerabaul29:
@AlexKuhnle I improved a bit to get a graceful exit, and I added Readme and example of use.

jerabaul29:
PS: I do not now if you want to add some licensing at the top of each file, or if the global license of the repo is enough? I am happy to cede the copyright as long as 1) this is distributed as full open source 2) our work is mentioned in the readme of the socket folder and / or at the beginning of the files.

AlexKuhnle:
Regarding license: It will be part of the [Apache 2 license](https://github.com/tensorforce/tensorforce/blob/master/LICENSE) and potentially the [usual header](https://github.com/tensorforce/tensorforce/blob/master/tensorforce/__init__.py). I'm of course happy to acknowledge the feature contribution, and I think the best place would be [here](https://github.com/tensorforce/tensorforce#core-team-and-contributors) (in whatever form you want), plus maybe in the class docstring. However, while I'm happy to keep it in a contrib folder for the start, I will work on integrating the feature into the framework and in the course likely reorganise things to some degree, mainly to avoid an unmaintained contrib "graveyard" with old code (as we had at an earlier stage). :-)  I hope you're fine with that? Acknowledgement and paper reference stay, of course, and because of that I think it's well-placed in the readme (could even include bibtex).

jerabaul29:
This is perfect to me :)

jerabaul29:
PS: I see the checks fail. Not sure what / why etc. Let me know if I can help with something.

jerabaul29:
Ok. I can confirm it works well on my laptop (Ubuntu 18.04) following both methods in the Readme. Tell me if you see any problem. Feel very free to clean / improve on the code ;)

jerabaul29:
PS: @AlexKuhnle anything more you want me to do on this? :)

AlexKuhnle:
Oh sorry, I forgot about it these last days. No, all fine, will merge it.

jerabaul29:
No worries :) Thanks :) If it is ok with you, I will be happy to be mentioned as contributor, but this is no worries.

AlexKuhnle:
Hi @jerabaul29, just to let you know that I finally managed to integrate this feature in a nice way (I think), and also combined it with the `multiprocessing` feature of vbelus. Environments can now be started as remote server via `Environment.create(..., remote='socket-server', port=65432)` and as local client proxy via `Environment.create(..., remote='socket-client', host='???', port=65432)`. It would be great to test it out.

jerabaul29:
@AlexKuhnle just to make it completely clear / sure:

Is it right that it is easy to switch between the 3 following modes using separate syntax? It is important to be able to use each of these modes, as depending on the exact use case (relative time spent in env vs. agent in particular), different methods may be better:

  - not using any multiprocessing / socket (i.e. just calling different separate threads or whatsoever was already in tf by default)

  - using multiprocessing

  - using socket

Can you confirm that this is the case, and just show a very short example of syntax for each of these?

AlexKuhnle:
Yes, and I will add an example later. Basically, `Runner` works more or less as before, if you use it as before, in "single execution mode". If you use the `remote` argument, the environment is wrapped based on either `multiprocessing` (what vbelus had done, I think you're aware of this, right?) or `socket` (as what happened in the parallelization version of this PR). You're right, depending on what the characteristics of the simulation etc are, different setups are preferable, and that was the main motivation to get it all in a shared interface.

AlexKuhnle:
`sync_timesteps` and `sync_episodes` also works as before, another thing that's new is, first, the `blocking` option (so that the runner waits for environment return and doesn't process opportunistically what's ready) and, more importantly, the `batch_agent_calls` option, which batches all environment execution returns in one call to the agent (to minimize the TensorFlow call overhead).

jerabaul29:
Sounds good. Yes, I am well aware of the work of Belus.

Ok, so I understand how the multiprocessing and socket approaches work, and I guess we can still use several envs without the ```remote``` argument to perform parallel training but 'directly' with the env, right? :)

Sounds excellent regarding the syncing options.

AlexKuhnle:
Yes, you can run multiple environments as before (and you could have your own custom communication wrapper applied to them), but if you use the `remote` argument, this is done for you based on either multiprocessing or socket. `multiprocessing` starts the environment in a separate process and returns the wrapper env communicating with this process, whereas `socket-client` starts a socket listener environment which requires a corresponding `socket-server` environment running somewhere. In both cases, correctly closing the environment, error handling, etc is taken care of, so things should work smoothly and without much user code necessary. I'll add examples to the doc soon.

jerabaul29:
This sounds perfect. I can test when I have time. Let me know when the doc is up.

AlexKuhnle:
Hmm, I thought I had posted it here already, but seems not the case: Here some first minimal instruction on how to use it: https://tensorforce.readthedocs.io/en/latest/basics/features.html
You can try to use the `run.py` script and pass the path of the environment, like `--environment module_py.YourEnvironment`. Alternatively, in code:

```python
# Environment machine 1
Environment.create(..., remote='socket-server', port=65432)

# Environment machine 2
Environment.create(..., remote='socket-server', port=65433)

# Agent machine
runner = Runner(
    agent=..., num_parallel=2, remote='socket-client',
    host=['host1', 'host2'], port=[65432, 65433]
)
runner.run(num_episodes=...)
```

jerabaul29:
Ok, looks good :)

I have not had the time to test (quite crazy weeks here), but I think the code looks good (gave it a detailed look). You have tested that everything works well with cartpole too, right? :)

Then I think everything is good for my part. If there are problems I let you know.

PR #630, opened 2019-11-01
Deconvolution

vbelus:
- Added deconvolutions (actually transposed convolutions) in `convolution.py`

- Changed the method to calculate the output shape of convolutions `get_output_spec`  so that it uses functions from tensorflow, which takes the dilation into account

### DIFF
--- a/tensorforce/core/layers/convolution.py
+++ b/tensorforce/core/layers/convolution.py
@@ -16,6 +16,7 @@
 from math import ceil
 
 import tensorflow as tf
+from tensorflow.python.keras.utils.conv_utils import deconv_output_length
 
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Many thanks for the PR, looks all good!

AlexKuhnle:
I've spotted 1-2 minor problems (like conv1d_transpose not actually using a transposed version, or not adding the layer to the dict so that it's accessible via specification key), which I fixed on the tf2 branch. The reason why it's on tf2 is that apparently only more recently they made the conv-transpose API more consistent (including adding a transpose version of conv1d, so no wonder you didn't use it :-). Anyway, just to let you know that there will be minor changes.

PR #634, opened 2019-11-12
Multiprocessing wrapper

vbelus:
This is a wrapper for multicore parallel training using multiprocessing.
To use it with the parallel runner, simply do as you would have before, but wrapping the environments with `ProcessWrapper`

An usage example:
```python3
from tensorforce.execution import ParallelRunner
from tensorforce.environments.environment_process_wrapper import ProcessWrapper

n_env = 8
envs = []

# # Without the wrapper
# for _ in range(n_env):
#     envs.append(Environment.create(...))

# With the wrapper
for _ in range(n_env):
    envs.append(ProcessWrapper(Environment.create(...)))

# Initialize the agent
agent = ...
agent.initialize()

# Initialize the runner
runner = ParallelRunner(agent=agent, environments=envs)

# Start the runner
runner.run(num_episodes=100, sync_episodes=False)
runner.close()
```

On a custom gym environment of mine (the one I used in this paper https://arxiv.org/abs/1910.07788), this wrapper speeds up the execution by a factor of 2.2 as you can see here, with 16 environments.
comparison_threading_multiprocessing

The reason it's not linearly increasing with the number of environments is:
- I only have 4 cores on my CPU
- the only multi-core computations are the executions in the environments

This means that with simple environments which doesn't require a lot of computation to execute one step, this will not necessarily be faster - actually, `CartPole-v0` is slightly faster to compute without the multiprocessing wrapper because of multiprocessing overhead.

### DIFF
--- a/tensorforce/environments/environment_process_wrapper.py
+++ b/tensorforce/environments/environment_process_wrapper.py
@@ -1,0 +1,72 @@
+from tensorforce.environments import Environment
+from tensorforce import TensorforceError, util
+
+from multiprocessing import Process
+from multiprocessing import Pipe
+
+import importlib
+import json
+import os
+from threading import Thread
+
+def worker(environment, conn2):
+    while True:
+        # Receive the environment method's name, and (optional) arguments
+        (name, *args, kwargs) = conn2.recv()
+
+        attr = object.__getattribute__(environment, name)
+        if hasattr(attr, '__call__'):
+            # Get what is returned by the method
+            result = attr(*args, **kwargs)
+        else:
+            # Get the attribute
+            result = attr
+
+        # Send this to the Wrapper
+        conn2.send(result)
+
+class ProcessWrapper(Environment):
+    def __init__(self, environment):
+        super(ProcessWrapper, self).__init__()
+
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Thanks very much for the PR, @vbelus. There will probably be changes to the code over time, in particular to make things coherent with [this feature](https://github.com/tensorforce/tensorforce/tree/master/tensorforce/contrib) -- I hope you don't mind that. :-)

vbelus:
I don't mind at all !

AlexKuhnle:
Hey @vbelus, thanks again for the contribution. I've managed to integrate the feature now, so that environments can now be wrapped in a multiprocessing "layer" via `Environment.create(..., remote='multiprocessing')` (and `ParallelRunner` features were merged with `Runner`). If you're still working with it, it would be greatly appreciated if you could test whether performance still behaves as before (and/or if anything else is not working as expected, of course). :-)

AlexKuhnle:
By the way, I will add an acknowledgement in the new file for what the new implementation is based on (your original implementation). If you want it to be mentioned in a certain way (linked to a paper or repo or something), let me know.

PR #637, opened 2019-11-25
Replace tf.logging with Python logging module.

sibyjackgrove:
tf.logging was removed in TF 2.0. This commit fixes the error when running the quickstart.py which uses tf.logging.


AlexKuhnle:
Good catch, thanks! :-)

PR #638, opened 2019-12-11
dqn config

christianhidber:
I added in the benchmarks/configs folder the dqn configuration following the one used by tf-agents in their tutorial. 

To actually compare the 2 implementations, the Cartpole-v0 env should be used and the training time set to 20000 timesteps. The dqn config now also uses the network definition used (1-layer, 100 neurons) by tf-agents. 

It wasn't quite obvious to me if and how these settings should/could baked into the benchmark. So I'm not sure of this additional config is actually helpful or not.

### DIFF
--- a/benchmarks/configs/dqn.json
+++ b/benchmarks/configs/dqn.json
@@ -1,0 +1,21 @@
+{
+    "agent": "dqn",
+    "network": [
+        {
+            "type": "dense",
+            "size": 100,
+            "activation": "relu"
+        }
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
You mean the environment config? I think it's fine as it is, as we don't want to maintain super-exact comparability with tf-agents or so, for Tensorforce -- meaning that even if we run a slightly different environment configuration, the config should ideally perform well.

PR #644, opened 2020-01-16
Support for JSON-encoding specific numpy types

Woellchen:
Implemented a simple JSONEncoder to serialize numpy floating point
numbers, which doesn't work otherwise. This is necessary for saving
an agent. More cases can be added once they are identified not working.


Woellchen:
i will fix this later, no worries :)

AlexKuhnle:
Hey, no worries about the test, I think one of them is currently failing (waiting for a fix from my side). I will merge the PR soon, thanks for it! :+1:

PR #666, opened 2020-04-20
CARLAEnvironment an environment wrapper for CARLA Simulator

Luca96:
Hi @AlexKuhnle,
I've implement a tensorforce's environment wrapper for the [CARLA simulator](http://carla.org/). The environment is still in development: as my thesis advances more features will be  provided.

CARLA can be used to research about different aspects of autonomous driving systems, so I decided to make the `CARLAEnvironment` as customizable as possible. For example, uses can define their own state/action space as well the reward function.

As I said before, there's some more features to add (e.g. weather support). There is a known issue with Runner, meaning that this environment does not work properly with runners (at least for the moment).

Moreover, I've included a couple of example usages. One of them:
```python
import pygame

from tensorforce import Agent
from tensorforce.environments import CARLAEnvironment


def training_example(num_episodes: int, max_episode_timesteps: int):
    env = CARLAEnvironment(debug=True)

    # Create your own agent
    agent = Agent.create(agent='ppo',
                         environment=env,
                         max_episode_timesteps=max_episode_timesteps,
                         batch_size=1)

    # Training loop
    env.train(agent=agent, 
              num_episodes=num_episodes, max_episode_timesteps=max_episode_timesteps, 
              weights_dir=None, record_dir=None)  

    pygame.quit()
```

Let me know for any issue about the code, design, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ Note that in general the replication is not 100% faithful, since the models as d
 - [OpenSim](http://osim-rl.stanford.edu/), reinforcement learning with musculoskeletal models.
 - [PyGame Learning Environment](https://github.com/ntasfi/PyGame-Learning-Environment/), learning environment which allows a quick start to Reinforcement Learning in Python.
 - [ViZDoom](https://github.com/mwydmuch/ViZDoom), allows developing AI bots that play Doom using only the visual information.
-
+- [CARLA](https://github.com/carla-simulator/carla), is an open-source simulator for autonomous driving research.
 
 
 ## Support, feedback and donating
@@ -179,6 +179,8 @@ Tensorforce is currently developed and maintained by [Alexander Kuhnle](https://
 Earlier versions of Tensorforce (<= 0.4.2) were developed by [Michael Schaarschmidt](https://github.com/michaelschaarschmidt), [Alexander Kuhnle](https://github.com/AlexKuhnle) and [Kai Fricke](https://github.com/krfricke).
 
 The advanced parallel execution functionality was originally contributed by Jean Rabault (@jerabaul29) and Vincent Belus (@vbelus). Moreover, the pretraining feature was largely developed in collaboration with Hongwei Tang (@thw1021) and Jean Rabault (@jerabaul29).
+
+The CARLA environment wrapper is currently developed by Luca Anzalone (@luca96).
 
 We are very grateful for our open-source contributors (listed according to Github, updated periodically):
 
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -32,5 +32,6 @@
     - World's map loading (e.g. random, specific, etc.)
     - Weather support
     - Pretraining and Free play (e.g. for data collection)
+    - State space with a temporal component.
     - ...
 9. To be determined...
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -27,4 +27,10 @@
     - Sequence states/actions
     - State-dependent actions
     - Conditional/hierarchical actions
-8. To be determined...
+8. CARLA environment
+    - Docs and assertions 
+    - World's map loading (e.g. random, specific, etc.)
+    - Weather support
+    - Pretraining and Free play (e.g. for data collection)
+    - ...
+9. To be determined...
--- a/examples/carla_env_examples.py
+++ b/examples/carla_env_examples.py
@@ -1,0 +1,91 @@
+"""A collection of examples for CARLAEnvironment"""
+
+import pygame
+
+from tensorforce import Agent
+from tensorforce.environments import CARLAEnvironment
+
+
+def training_example(num_episodes: int, max_episode_timesteps: int):
+    # Instantiate the environment (run the CARLA simulator before doing this!)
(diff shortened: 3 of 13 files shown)


AlexKuhnle:
Sorry for the late response. :-/ 

Thanks very much for the PR, that sounds great, particularly also the example script (the examples folder really shouldn't be as empty as it currently is). I will have a look at it over the weekend. Please also feel free to add acknowledgements as you see fit, e.g. [here](https://github.com/tensorforce/tensorforce#core-team-and-contributors), particularly if you plan to extend it and keep it updated (which I would obvs encourage :-).

AlexKuhnle:
Very nice and well-integrated, thanks again. I haven't run it yet, but I will just leave this feature to you -- as I said, feel free to add information about your contribution (and probably maintenance/future development) somewhere.

One thing which will need changing: Currently, the way imports are structured, the framework will break for users who don't satisfy the CARLA requirements. Since Tensorforce is not environment-focused, a user shouldn't be required to install any env requirements. I've worked around this by moving the env-specific imports into the constructor and where else they are needed, see e.g. [here](https://github.com/tensorforce/tensorforce/blob/master/tensorforce/environments/openai_gym.py#L143). Since the `carla` subfolder is not imported anywhere else, the only file that needs changing is the `carla_environment.py, specifically:

```python
import carla
import pygame
# ...
from tensorforce.environments.carla import env_utils, CARLASyncContext
from tensorforce.environments.carla import Sensor, SensorSpecs
from tensorforce.environments.carla.env_utils import WAYPOINT_DICT
# ...
```

Could these be (re)moved from global imports into constructor/methods where they are used?

Another minor comment: You could add the dependencies, at least the Python ones, to the `setup.py` file as a setup option [here](https://github.com/tensorforce/tensorforce/blob/master/setup.py#L115).

AlexKuhnle:
Oh, and you can also add your environment (plus, for instance, acknowledgement) [here](https://github.com/tensorforce/tensorforce#environment-adapters).

Luca96:
I'm glad you like it :). As you suggested I have:
- Updated the [README](https://github.com/Luca96/tensorforce/blob/master/README.md) both "Environment adapters" and "contributors" sections.
- Included the required dependencies (`pygame` and `opencv`) in [setup.py](https://github.com/Luca96/tensorforce/blob/master/setup.py#L116), unfortunately there is no pip package for `carla` so users should follow the instructions provided in the CARLAEnvironment's [docs](https://github.com/Luca96/tensorforce/blob/master/tensorforce/environments/carla_environment.py#L29-L42).
- Written some notes about future developments in [ROADMAP](https://github.com/Luca96/tensorforce/blob/master/ROADMAP.md).

I think that the above code should fix the [import issue](https://github.com/Luca96/tensorforce/blob/master/tensorforce/environments/carla_environment.py#L12-L18) nicely:
```python
try:
    from tensorforce.environments.carla import env_utils
    from tensorforce.environments.carla.env_utils import WAYPOINT_DICT
    from tensorforce.environments.carla.sensors import Sensor, SensorSpecs
    from tensorforce.environments.carla.synchronous_mode import CARLASyncContext
except ImportError:
    pass
```
This also allows users who want to customize the env to import stuff from `carla_environment` instead from the forlder `carla`. Here is an [example](https://github.com/Luca96/tensorforce/blob/master/examples/carla_examples.py#L33) about that.

Let me know if there is some more to do. :wink:

AlexKuhnle:
Thanks, looks good. Last minor change: The `carla` and `pygame` import should also be within the try-except block (and should `pygame` be part of the `setup.py` imports?).

Luca96:
Thanks for pointing it out (I forgot them...) Anyway, what do you mean about `pygame` being part of `setup.py`? You lirerally mean `import pygame` on the very first lines? [This](https://github.com/Luca96/tensorforce/blob/master/setup.py#L116) isn't enough?

AlexKuhnle:
Yes, that's what I had in mind... but you're right, it's already included, so not sure what I was thinking. Does `carla` also need to be included? Otherwise just forget the comment. :-D

Luca96:
Alright ;D. Yes, `carla` should be included but we cannot do this because there is no pip package (actually, there's one but is a little bit outdated), unfortunately. Users should install it manually, as I wrote [here](https://github.com/Luca96/tensorforce/blob/master/tensorforce/environments/carla_environment.py#L29-L42) or ufficially stated [here](https://carla.readthedocs.io/en/latest/build_system/#pythonapi). I've just noticed that they released a new version 3 days ago which should improve the python's API installation. In fact, now they provide a [setup.py](https://github.com/carla-simulator/carla/blob/master/PythonAPI/carla/setup.py).

AlexKuhnle:
I see. Well, in this case forget about the setup and carla comment as well... :-)

Luca96:
I think is everything fixed now. I also made a change in [keras.py](https://github.com/tensorforce/tensorforce/pull/666/commits/96ae18b09fc85d913c1bf71f2326f2576bab8302#diff-eab5d338564121bc56f7146530ac6600).

AlexKuhnle:
All good, merging now. :-)

PR #671, opened 2020-04-27
Quickstart example of a off/on thermostat RL system

mpettis:
Per discussion, this is a self-contained Jupyter notebook example of making a RL agent to learn how to control a room temperature within a temperature band with a off/on heater.


AlexKuhnle:
Thanks very much, PRs like this are very welcome.

AlexKuhnle:
FYI: I copy-pasted your example code as a unit-test (plus reduced episode number), to make sure it doesn't break:  https://github.com/tensorforce/tensorforce/blob/master/test/test_examples.py#L68

PR #688, opened 2020-06-08
corrected filename truncation for h5-files

afaul:
The filename truncation should be different, depending if the filename ends with ".hdf5" or ".h5"


AlexKuhnle:
Good catch, thanks very much! :-)

PR #695, opened 2020-06-12
Save checkpoint instead of write

bezineb5:
Hi,

I switched the checkpoint saving method from "write" to "save", as write doesn't create a "checkpoint" file which allows restoring afterwards (see https://www.tensorflow.org/api_docs/python/tf/train/Checkpoint#write). Was it intentional to use write?

I also added an error message when there's no checkpoint, and simplified the code to extract the filename. Nothing really fancy.

Thanks!
Benjamin

### DIFF
--- a/tensorforce/agents/agent.py
+++ b/tensorforce/agents/agent.py
@@ -1054,11 +1054,9 @@ class Agent(object):
                     assert format == 'checkpoint'
                 if filename is None or not os.path.isfile(os.path.join(directory, filename + '.index')):
                     path = tf.train.latest_checkpoint(checkpoint_dir=directory)
-                    # TODO: path or filename returned?
-                    if '/' in path:
-                        filename = path[path.rindex('/') + 1:]
-                    else:
-                        filename = path
+                    if not path:
+                        raise TensorforceError.exists_not(name='Checkpoint', value=directory)
+                    filename = os.path.basename(path)
 
             else:
                 if format is None:
--- a/tensorforce/core/models/model.py
+++ b/tensorforce/core/models/model.py
@@ -688,7 +688,13 @@ class Model(Module):
             # always write temporary terminal=2/3 to indicate it is in process... has been removed recently...
             # check everywhere temrinal is checked that this is correct, if 3 is used.
             # Reset should reset estimator!!!
-            return super().save(directory=directory, filename=filename)
+            if self.checkpoint is None:
+                self.checkpoint = tf.train.Checkpoint(**{self.name: self})
+            
+            # We are using the high-level "save" method of the checkpoint to write a "checkpoint" file.
+            # This makes it easily restorable later on.
+            # The base class uses the lower level "write" method, which doesn't provide such niceties.
+            return self.checkpoint.save(file_prefix=os.path.join(directory, filename))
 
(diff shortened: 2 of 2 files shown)


AlexKuhnle:
I will have to look into that later. The tests have been working fine with the current version, so need to check why they didn't complain. Thanks for pointing it out and for the PR!

AlexKuhnle:
Had a look now and remember why `save` wasn't used: On the one hand, it uses its own internal counter, on the other hand, "because" it creates the checkpoint file. However, the reason was mainly with respect to general modules as low-level save functionality (say, you want to save your network or so), which is not considered to result in a functional model or so. You're right that for the special case of the `Model` `Module`, this should behave differently. So `Model.save()` should not call `super`, but handle it differently. Ideally, there would be a `CheckpointManager`-like functionality, so that custom `append` arguments are handled accordingly -- have to think more about it. Would you mind reverting the `write` change in `Module` and instead copy-paste it [here](https://github.com/tensorforce/tensorforce/blob/tf2/tensorforce/core/models/model.py#L691) (and remove the `super` call)? That should do it, I guess...

bezineb5:
Indeed, that makes sense - I'll move it to the model.

Also, I was looking at the saved_model export. You mentioned somewhere that you get an exception referencing tensorforce.python.util.nest._DotString (I don't have the exact exception at hand). I happens basically when trying to print an error message when tensorflow checks the signature, so in itself, it's not import. When fixed, it gives this error:
```
ValueError: in user code:

    /usr/local/lib/python3.7/site-packages/tensorflow/python/saved_model/signature_serialization.py:126 signature_wrapper  *
        structured_outputs, signature_function.name, signature_key)
    /usr/local/lib/python3.7/site-packages/tensorflow/python/saved_model/signature_serialization.py:188 _normalize_outputs  **
        .format(original_outputs, function_name, signature_key))

    ValueError: Got non-flat outputs '(TensorDict(heading=Tensor("StatefulPartitionedCall:0", shape=(None,), dtype=float32)), TensorDict())' from 'b'__inference_function_graph_1731'' for SavedModel signature 'serving_default'. Signatures have one Tensor per output, so to have predictable names Python functions used to generate these signatures should avoid outputting Tensors in nested structures.
```

I guess that the output should be flat. I could help on that but I fear that it might induce a lot of changes. Any thought about this?

Thanks a lot!

AlexKuhnle:
Yes, I suspected at some point that this DotString message was because it got entangled with an exception while tracing the graph, or so. However, it wasn't obvious what exactly needs to be changed to get to the "true" exception -- how did you do it?

AlexKuhnle:
And so you're saying the reason is that every function has to return flattened outputs, or just the outer-most saved function object? Interesting, and makes kind of sense... I was wondering why `tf.function` flattens inputs but returns nested outputs -- apparently something one shouldn't rely on. Well, it may not be all that problematic, since this whole `tf.function` wrapping inside Tensorforce likely requires a bit more attention at some point. I have some ideas around that, and if you want to get involved on that, you're very welcome and we can discuss more -- it will get you rather deep into internals, though. :-D

AlexKuhnle:
Anyway, PR looks good now, many thanks!

Issue #696, opened 2020-06-12
unbound actionspace in env result in nan actions


stheid:
I have an environment that implements the OpenAI Gym interface, hence i load it with the openai gym env wrapper.

when my env returns an unbounded actionspace
```python
self.action_space = gym.spaces.Box(low=np.full(d_i, -np.inf), high=np.full(d_i, np.inf))
```

```
Episodes:   0%|                                                                                                                                     | 0/10 [00:00, reward=0.00, ts/ep=0, sec/ep=0.00, ms/ts=0.0, agent=0.0%]Traceback (most recent call last):
  File "/home/sheid/sheid/openmodelica-microgrid-gym/experiments/PPO/tensorforce_exp/main.py", line 22, in <module>
    runner.run(num_episodes=10)
  File "/home/sheid/miniconda3/envs/openmodelica-microgrid-gym-py37/lib/python3.7/site-packages/tensorforce/execution/runner.py", line 469, in run
    observation = self.environments[n].receive_execute()
  File "/home/sheid/miniconda3/envs/openmodelica-microgrid-gym-py37/lib/python3.7/site-packages/tensorforce/environments/environment.py", line 325, in receive_execute
    states, terminal, reward = self.execute(actions=self._actions)
  File "/home/sheid/miniconda3/envs/openmodelica-microgrid-gym-py37/lib/python3.7/site-packages/tensorforce/environments/environment.py", line 366, in execute
    states, terminal, reward = self.environment.execute(actions=actions)
  File "/home/sheid/miniconda3/envs/openmodelica-microgrid-gym-py37/lib/python3.7/site-packages/tensorforce/environments/openai_gym.py", line 220, in execute
    actions = OpenAIGym.unflatten_action(action=actions)
  File "/home/sheid/miniconda3/envs/openmodelica-microgrid-gym-py37/lib/python3.7/site-packages/tensorforce/environments/openai_gym.py", line 379, in unflatten_action
    raise TensorforceError("Action contains inf or nan.")
tensorforce.exception.TensorforceError: Action contains inf or nan.
```

## Workaround

when my env returns a bounded actionspace
```python
self.action_space = gym.spaces.Box(low=np.full(d_i, -10e24), high=np.full(d_i, 10e24))
```
everything seems to run fine

stheid:
Unbounded observation spaces however cause no issues!

```python
self.observation_space = gym.spaces.Box(low=np.full(d_o, -np.inf), high=np.full(d_o, np.inf))
[...]

PR #698, opened 2020-06-15
Update setup.py

stheid:
minimal tensorflow version 2.2

i think its worth validating that again, but it seems (#696) the new branch does **not** work with:
- tf 2.0.1
- tf 2.1.1

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -103,7 +103,7 @@ setup(
     python_requires='>=3.5',
     install_requires=install_requires,
     extras_require=dict(
-        tf=['tensorflow'],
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Thanks. I will look into that...

AlexKuhnle:
This was due to 1-2 minor issues, but should now be fixed (at least for `2.0.1`). Feel free to correct to `>=2.0.1` or close it.

AlexKuhnle:
Sorry, I realize that it may still be necessary...

PR #745, opened 2020-11-06
Unpacking Experience Fix for Remote Environments.

tahaeghtesad:
Fixed an issue with unpacking experiences (e.g., states, terminal, reward, seconds) from a remote environment.

Without unpacking the seconds, which is added by each worker environment, [this assertion](https://github.com/tensorforce/tensorforce/blob/67424fcc2f2c8d92dce6f163fad1fbe0c338fa1d/tensorforce/agents/recorder.py#L502) is failed since seconds will be a part of states.

I was intending to raise an issue but debugging was faster

### DIFF
--- a/tensorforce/environments/environment.py
+++ b/tensorforce/environments/environment.py
@@ -650,9 +650,12 @@ class RemoteEnvironment(Environment):
     def receive_execute(self):
         if self._blocking:
             if self._expect_receive == 'reset':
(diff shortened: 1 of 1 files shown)


AlexKuhnle:
Many thanks, looks good! And no worries, PRs are as welcome as issues... :-)  Need to check why this didn't show up in the unittests.

Issue #799, opened 2021-06-12
Error with the examples on latest version installation 


saleh-mir:
Hi there and thank you for the fantastic framework. 

I just installed everything on my mac M1 machine using conda. I'm using Tensorflow `2.5`. After I run the example of repo (or tried another one about custom ENVs) I get this error:

```
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-1-384aca331f84> in <module>
      7 
      8 # Instantiate a Tensorforce agent
----> 9 agent = Agent.create(
     10     agent='tensorforce',
     11     environment=environment,  # alternatively: states, actions, (max_episode_timesteps)

~/miniforge3/lib/python3.9/site-packages/tensorforce/agents/agent.py in create(agent, environment, **kwargs)
    114                 # Keyword specification
    115                 agent = tensorforce.agents.agents[agent]
--> 116                 return Agent.create(agent=agent, environment=environment, **kwargs)
    117 
    118             else:

~/miniforge3/lib/python3.9/site-packages/tensorforce/agents/agent.py in create(agent, environment, **kwargs)
     87             agent = agent(**kwargs)
     88             assert isinstance(agent, Agent)
---> 89             return Agent.create(agent=agent, environment=environment)
     90 
     91         elif isinstance(agent, dict):

~/miniforge3/lib/python3.9/site-packages/tensorforce/agents/agent.py in create(agent, environment, **kwargs)
     64                 agent.reset()
     65             else:
---> 66                 agent.initialize()
     67 
     68             return agent

~/miniforge3/lib/python3.9/site-packages/tensorforce/agents/agent.py in initialize(self)
    355             raise TensorforceError(message="Missing agent attribute model.")
    356 
--> 357         self.model.initialize()
    358 
    359         self.internals_spec = self.model.internals_spec

~/miniforge3/lib/python3.9/site-packages/tensorforce/core/models/model.py in initialize(self)
    285                 quit()
    286 
--> 287         super().initialize()
    288 
    289         # If we are a global model -> return here.

~/miniforge3/lib/python3.9/site-packages/tensorforce/core/module.py in initialize(self)
    544                     raise TensorforceError.unexpected()
    545 
--> 546                 function = self.create_api_function(
    547                     name='{}.{}'.format(self.name, function_name), api_function=api_function
    548                 )

~/miniforge3/lib/python3.9/site-packages/tensorforce/core/module.py in create_api_function(self, name, api_function)
[...]

PR #800, opened 2021-06-13
Add installation notes for M1 Macs

saleh-mir:
As discussed in #799 . 

Please let me know if I need to change it.


AlexKuhnle:
Thank you!

PR #809, opened 2021-07-06
Fixes for Keras network compatibility

Tianyue-Zhao:
A fix for compatibility with the Keras network wrapper and Tensorboard summaries :)


AlexKuhnle:
Thanks for the PR!

Issue #831, opened 2021-09-30
Bugs to do with experience


ishanpai:
While exploring TensorForce with @fkiraly-shell, I came across what I think are two bugs in TensorForce. They were both to do with the experience method of an agent. I have created a notebook with a reproducible example of all the problems I describe. I have attached this notebook in a draft pull request #832.

I am using python 3.9.7 and Tensorforce 0.6.5 and pandas 1.3.3. My environment has no other packages installed separately.

**Problem 1**
The first problem is that `experience` does not seem to work in "batch" mode. We have the following set up:
```python
import numpy as np
import pandas as pd
from tensorforce.agents import Agent
np.random.seed(0)
n_train = 10
X_train = pd.DataFrame(np.random.uniform(size=n_train), columns=["A"])
Z_train =  pd.DataFrame(np.array([np.random.uniform(size=n_train), np.random.uniform(size=n_train)]).T, columns=["B", "C"])
Y_train = pd.DataFrame(np.random.uniform(size=n_train), columns=["D"])

agent = Agent.create(
    agent='tensorforce', states=dict(type='float', shape=(2,), min_value=(0,0), max_value=(1,1)),
    actions=dict(type='float', shape=(1,), min_value=0, max_value=1),
    max_episode_timesteps=2, update=64,
    optimizer=dict(optimizer='adam', learning_rate=1e-3),
    objective='policy_gradient', reward_estimation=dict(horizon=1),
    exploration=10.0, parallel_interactions=n_train
)
```

Here, `X_train`, `Z_train`, `Y_train` are all `pd.DataFrame`.
I tried batch mode with the following code:

```python
states = [{"state": np.array(Z_train.iloc[i])} for i in range(len(Z_train))]
actions = [{"action": np.array(X_train.iloc[i])} for i in range(len(X_train))]
rewards = np.reshape(np.array(Y_train), (-1,))
terminal = np.full(rewards.shape, True)

agent.experience(states=states, actions=actions, terminal=terminal, reward=rewards)
```

- This runs into the following error: `TensorforceError: Invalid value for Agent.experience actions argument value: ArrayDict(action=[[0.54881352]]).` which suggests that `experience` does not actually want an array of dictionaries for `actions`, while it does seem to want this for `states`.
[...]

PR #832, opened 2021-09-30
added notebook with bugs

ishanpai:
Notebook with detailed code of when the bugs in #831 appear, showing the corresponding error messages.


AlexKuhnle:
Let me know if the problem is solved, so this PR can be closed.

ishanpai:
Yes, the problem seem to be solved, sorry for the late response.

PR #849, opened 2021-12-28
make states ArrayDict in case _states is a nested dict.

dxaen:
The assertion contained in `_check_states_output()` fails if `_states` is nested since the `len()` operator only returns the high-level value for `dict()`.


AlexKuhnle:
Looks good, merging.

dxaen:
Thanks @AlexKuhnle

PR #899, opened 2023-08-14
Set maximum `gym` version for compatibility reasons.

bwarre471:
With higher gym versions, it shows :

`AttributeError: module 'gym.wrappers' has no attribute 'Monitor'`

I attempt to fix it, but it throwed further errors that makes it non working.
In some versions, it's error about `gym.envs.registry.env_specs` that needs to be changed to `gym.envs.registry` in whole `openai_gym.py` file, but there was also errors about `assert spec is not None` and it was done, as API to specs changed.

### DIFF
--- a/requirements-all.txt
+++ b/requirements-all.txt
@@ -1,4 +1,4 @@
-gym >= 0.21.0
+gym >= 0.21.0, <0.23
 h5py >= 3.6.0
 matplotlib >= 3.5.1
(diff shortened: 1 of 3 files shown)


bwarre471:
I don't know if that kwargs thing won't broke it, but like that it works on
my machine and otherwise it throws error making an environment.

